PR #3, opened 2018-02-02
Create signal handler to respond when BlockCompletions are updated 

jcdyer:
**Description:** This connects the completion_aggregator app to the signal fired when a BlockCompletion gets saved.  A follow-up PR (#4) will implement responding to that signal to update Aggregators for the course.  For now, it just logs that the update happened.

**JIRA:** MCKIN-5902, OC-3097

**Dependencies:** Depends on aggregator model PR, #1, currently waiting on upstream review.

**Merge deadline:** None

**Testing instructions:** 

1. Run the test suite, and see that tests pass.
2. Configure logging to display DEBUG level logs.
3. Install into a current version of edx-platform. (pip install the repo, at this branch and update the INSTALLED_APPS in lms/envs/common.py)
4. Enable the completion waffle switch. (completion.enable_completion_tracking)
5. Work through a course to create new completion objects, or manually create BlockCompletions from the shell.
6. See that the logs were updated with entries from the completion_aggregator namespace.

**Reviewers:**
- [ ] @macornwell
- [ ] @iloveagent57

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added
- [ ] Documentation updated (not only docstrings)
- [ ] Commits are squashed
- [ ] PR author is listed in AUTHORS

**Post merge:**
- [ ] Delete working branch (if not needed anymore)

**Author concerns:** Code isn't really a complete feature as-is.  Not useful until the follow-up PR lands.

PR #2, opened 2018-01-04
Aggregate Completion Views

jcdyer:
This is moved over from edx/edx-platform#17015
# OC-3096  Aggregate Completion Views (WIP)

**Description:**  Moved PR to new repo.  

This still has some hard links to edx-platform, which are keeping the tests from running.

**JIRA:** 
*   OC-3096 - Completion views 
*   OC-3571 - Move completion aggregator to separate repo.

## Original Description

Code is adapted from the solutions branch v0 api.  Currently has detail and list views working, but not block update view.

Some issues still exist around handling missing completion records, but those won't be resolved until the actual calculation of AggregateCompletion blocks is implemented (see comments on expectedFailure tickets).

The branch is based on cornwell/create-model  (#16715).  Once that merges, this will be re-pointed to edx:master.

### Test directions

- Manually create AggregateCompletion objects with aggregation_name `'course'`, and others with the same course_key and user, with an aggregation_name of `'chapter'`, `'sequential'`, or `'vertical'`.
- Visit /api/completion/v1/course/{course_key}/, and verify that the course record shows up.
- Add the querystring ?requested_fields=chapter,sequential,vertical (or some combination of those), and verify that the other AggregateCompletion records show up.
- Create some AggregateCompletion records in another course.
- Visit /api/completion/v1/course/?requested_fields=chapter,sequential,vertical and verify that the aggregate completions for both courses show up in the appropriate structure.

### Reviewers

- [x] @macornwell
- [ ] @iloveagent57 

FYI: @bradenmacdonald

PR #4, opened 2018-02-16
Update aggregators whenever a block completion is updated.

jcdyer:
**Description:** This adds the logic to update aggregators whenever a new block completion is created.

Manual test instructions are a little complex.  I'll update them in the next hour or so.

**Merge deadline:** None

1.  Checkout edx/edx-platform to the cliff/aggregator-settings branch
2.  Run paver install_python_prereqs

**Testing instructions:**

1.  Set the completion.enable_completion_tracking waffle switch.
2.  Delete existing BlockCompletion objects:

        >>> from completion.models import BlockCompletion
        >>> BlockCompletion.objects.all().delete()

3.  Visit some pages in a course.
4.  Check that completion objects were created.

        >>> BlockCompletion.objects.all()

5. Check that Aggregators were created for course, chapter, sequential, and vertical blocks, and have sensible values.

        >>> from completion_aggregator.models import Aggregator
        >>> Aggregator.objects.all()

**Reviewers:**
- [x] @macornwell 
- [x] @iloveagent57

FYI: @bradenmacdonald 

**Merge checklist:**
- [x] All reviewers approved
- [ ] CI build is green
- [x] Documentation updated
- [x] Commits are squashed
- [x] PR author is listed in AUTHORS

**Post merge:**

**Author concerns:**


.travis.yml
```diff
@@ -8,15 +8,16 @@ python:
...
+  - TOXENV=django111
 
 matrix:
   include:
     - python: 3.5
       env: TOXENV=quality
     - python: 3.5
       env: TOXENV=docs
+    - python: 3.5
+      env: TOXENV=django20
```
jcdyer: Tangential change: edx is jumping from django 1.8 to django 1.11, so there's no reason to test against 1.9 and 1.10.  Django 2.0 came out so test it too, but it only works with python 3, not python2.

completion_aggregator/models.py
```diff
@@ -145,6 +145,7 @@ def submit_completion(self, user, course_key, block_key, aggregation_name, earne
                 'percent': percent,
                 'possible': possible,
                 'earned': earned,
+                'last_modified': last_modified,
```
jcdyer: I added this field (even though there's an auto-updated "modified" field, because the way that aggregators are updated from block completions means that relying on the actual time it was updated introduces race conditions that could cause completions to get missed.  `last_modified` is updated to the timestamp of the most recently seen block completion, so any newer objects will have a newer timestamp & there's no race condition.

macornwell:
@jcdyer  the most micro of changes. Under the PR description you have 
```
from completion_aggregators.models import Aggregator
Aggregator.objects.all()
```
It should be:
`from completion_aggregator.models import Aggregator `
(no s in aggregator)

macornwell:
@jcdyer  everything looks good and works as expected. I ran `paver test_python`  and hit a few errors. I don't know if they're just on my side, but there isn't CI so I figured I'd check.

jcdyer:
Updated the PR description as requested.  CI is running, hidden behind "Show all checks" next to "All checks have passed" below.

jcdyer:
@macornwell This is ready for another review.

macornwell:
@jcdyer  you're right about the CI! Thanks for pointing that out. Looks good to me.

jcdyer:
Thanks @macornwell.  @iloveagent57: This is now RFR.  There is a sequence of PRs now pending edX review: #1 (just needs final confirmation, I believe), #3, and #4.  

#2 is also waiting on edX review, but is only dependent on #1.

I'd love to start getting some of these merged, as the chain of dependencies is making branch management difficult.

.travis.yml
jcdyer: @ciuin - Here's where I'm changing the tox environments to run in.  Feel free to use the same config on your branch, to keep merges easier.

completion_aggregator/models.py
iloveagent57: It's probably worth mentioning this in the docstring for this method, or maybe the model's docstring.

completion_aggregator/signals.py
```diff
@@ -31,13 +34,27 @@ def completion_update_handler(signal, sender, instance, created, raw, using, upd
...
     )
 
+    if not Aggregator.objects.filter(
+            user=instance.user,
+            course_key=instance.course_key,
+            aggregation_name='course',
+            last_modified__gte=instance.modified).exists():
+
+        try:
+            update_aggregators(
```
iloveagent57: does this need an `apply_async()`?

completion_aggregator/tasks.py
```diff
@@ -0,0 +1,188 @@
...
+        Return the list of BlockCompletions.
+
+        Each must have the following attributes:
+
+            .block_key (UsageKey)
+            .modified (datetime)
+            .completion
+        """
+        from completion.models import BlockCompletion  # pylint: disable=import-error
+        return BlockCompletion.objects.filter(
```
iloveagent57: There's a classmethod on the `BlockCompletion` model now that will do this for you and also give you the dictionary you construct in `__init__()`. https://github.com/edx/completion/blob/master/completion/models.py#L188

completion_aggregator/tasks.py
```diff
@@ -0,0 +1,188 @@
...
+            Aggregator.objects.submit_completion(
+                user=self.user,
+                course_key=self.course_key,
+                block_key=block,
+                aggregation_name=block.block_type,
+                earned=total_earned,
+                possible=total_possible,
+                last_modified=last_modified,
+            )
+        return (total_earned, total_possible, last_modified)
```
iloveagent57: These three methods all return the same tuple data structure, maybe it's worth creating a namedtuple?

completion_aggregator/tasks.py
```diff
@@ -0,0 +1,188 @@
...
+        self.block_completions = {completion.block_key: completion for completion in self._get_block_completions()}
+
+    def update(self, changed_blocks=frozenset()):
+        """
+        Update the aggregators for the course.
+
+        Takes a set of completable blocks that have been recently updated to
+        inform how to perform the update.  Currently no optimizations are
+        performed based on this information, but in the future they may help
+        cut down on the amount of work performed.
```
iloveagent57: Is the idea that, since you have the BlockStructure for the course, you can just search the structure for any node with a corresponding aggregator that has one of the `changed_blocks` in its subtree?

tox.ini
```diff
@@ -18,15 +18,14 @@ match-dir = (?!migrations)
...
+addopts = --cov completion_aggregator --cov-report term-missing --cov-report xml -W error
 norecursedirs = .* docs requirements
 
 [testenv]
 deps =
     django18: Django>=1.8,<1.9
-    django19: Django>=1.9,<1.10
-    django110: Django>=1.10,<1.11
     django111: Django>=1.11,<1.12
+    django20: Django>=2.0,<2.1
```
iloveagent57: Nice

completion_aggregator/signals.py
jcdyer: Yes.

completion_aggregator/tasks.py
jcdyer: Spiffy!  I'll put it to good use.

completion_aggregator/tasks.py
jcdyer: Yeah, that's pretty much it.  It seems that once you've traversed the course graph, summing the completions is pretty near free (computers are really good at addition), but if we can decide *before* traversing the tree that we trust a given aggregator's value, we can skip the traversal altogether.  Whether that's by calculating the descendents of an aggregator block or the ancestors of the completable blocks, I'm not sure.

completion_aggregator/tasks.py
jcdyer: That's a good idea.

jcdyer:
Nimisha had suggested doing a periodic task rather than updating the tree on each update.  We have a story in the backlog to implement that, but we're opting to get feature-completeness first, and then optimize once it's working.

jcdyer:
I'll push fixes on this in the morning.

completion_aggregator/tasks.py
jcdyer: Looking closer at this, I'd like to include this work in #6 instead.  Because it changes the abstraction barriers, introducing this here would make merging #6 significantly more painful.

jcdyer:
Ready for another review.  The django 1.11 and 2.0 tests are failing because of a new dependency on completion (Yay for that being outside edx-platform now!), and should be resolved when edx/completion#10 lands.  Please give that a quick review, too.

jcdyer:
Oops.  Accidentally closed the PR.  Reopened now.

completion_aggregator/tasks.py
iloveagent57: 👍 Gotchya

iloveagent57:
@jcdyer there are different `edx-completion` versions specified in doc.txt from test.txt

jcdyer:
Fixed.

PR #10, opened 2018-05-01
Add service to periodically aggregate out of date completions.

jcdyer:
**Description:** 

This provides a service that runs out-of-band and updates out-of-date completion aggregators.
When a completion would previously have called the aggregator task directly, it now marks its BlockCompletions out-of-date in a StaleCompletion table.  When the service runs, it gets all the stale blocks from the table, runs one completion task for each enrollment that is out of date, and marks all its stale completions resolved.  Later, resolved completions are deleted from the table, but we don't do this immediately to reduce table churn, and to provide a little visibility into recent table activity.

**JIRA:** OC-3968

**Dependencies:** None

**Merge deadline:** None

**Installation instructions:** None.

**Testing instructions:**
1. Run the management command.  ./manage.py run_aggregator_service -v2
2. Create some BlockCompletions, either by browsing a course, or from a ./manage.py shell
3. Verify (from the logging or from the command line) that StaleCompletions are being created and resolved.  
4. If running from edx-platform, verify that running a celery task worker causes the aggregators to actually be updated.  If running from the openedx-completion-aggregator repo, verify that running a celery task worker causes tasks to be taken off the queue, but fail, because the modulestore can't be accessed.

**Reviewers:**
- [ ] @pomegranited  

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**

**Author concerns:** We still need to make sure this service gets run from a cronjob when deployed to 
McKinsey.  I'm not sure how we do that.


completion_aggregator/aggregator.py
```diff
@@ -0,0 +1,173 @@
...
+                else:
+                    stale_blocks[enrollment].add(stale.block_key)
+                if stale.force:
+                    forced_updates.add(enrollment)
+
+            log.info("Performing aggregation update for %s user enrollments", len(stale_blocks))
+            for enrollment in stale_blocks:
+                if isinstance(stale_blocks[enrollment], utils.BagOfHolding):
+                    blocks = []
+                elif len(stale_blocks[enrollment]) > 128:
```
pomegranited: Please define this magic `128` number as a constant above.

completion_aggregator/aggregator.py
```diff
@@ -0,0 +1,173 @@
...
+                if stale.block_key is None:
+                    stale_blocks[enrollment] = utils.BagOfHolding()
+                else:
+                    stale_blocks[enrollment].add(stale.block_key)
+                if stale.force:
+                    forced_updates.add(enrollment)
+
+            log.info("Performing aggregation update for %s user enrollments", len(stale_blocks))
+            for enrollment in stale_blocks:
+                if isinstance(stale_blocks[enrollment], utils.BagOfHolding):
```
pomegranited: I really want to combine this condition with the next one by making `len(BagOfHolding()) == float('inf')`..  But `__len__()` must return an integer, and infinity won't fit in a lowly integer 😦

completion_aggregator/aggregator.py
```diff
@@ -0,0 +1,173 @@
...
+                    blocks = []
+                else:
+                    blocks = [six.text_type(block_key) for block_key in stale_blocks[enrollment]]
+                tasks.update_aggregators.delay(
+                    username=enrollment.username,
+                    course_key=six.text_type(enrollment.course_key),
+                    block_keys=blocks,
+                    force=enrollment in forced_updates,
+                )
+            models.StaleCompletion.objects.filter(id__in=(stale.id for stale in stale_queryset)).update(resolved=True)
```
pomegranited: Does work instead?
```python
stale_queryset.update(resolved=True)
```

completion_aggregator/aggregator.py
```diff
@@ -0,0 +1,173 @@
...
+        """
+        Start the service.
+        """
+        service_start_time = timezone.now()
+        service_end_time = service_start_time + self.runtime
+        tick = service_start_time
+        i = 0
+        while tick < service_end_time:
+            i += 1
+            if i % 1000 == 0:
```
pomegranited: Seems arbitrary to run cleanup every 1000 ticks?

completion_aggregator/signals.py
```diff
@@ -56,7 +53,7 @@ def item_deleted_handler(usage_key, user_id, **kwargs):
     # only fired from split-mongo, so it will always contain the course run.
     course_key = usage_key.course_key
     for user in get_active_users(course_key):
-        update_aggregators.delay(username=user.username, course_key=six.text_type(course_key), force=True)
+        StaleCompletion.objects.create(username=user.username, course_key=course_key, force=True)
```
pomegranited: Does it make sense to add a waffle switch to decide whether to process completions with the original celery task, or to mark them Stale and process in batch?

completion_aggregator/aggregator.py
```diff
@@ -0,0 +1,173 @@
...
+
+    def __init__(self, runtime=None, period=None):
+        """
+        Create a PeriodicService.
+
+        **Arguments:**
+
+        runtime (timedelta|None):
+            Run the service for this length of time, then quit.  If not
+            provided, run forever.
```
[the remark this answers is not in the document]
jcdyer: I was thinking of it as a fallback to the service failing, but I suppose running it under supervisor would be a reasonable way to make sure it keeps running.

As for not running it under celerybeat, I was remembering some conversations with @jibsheet, where he said he didn't really like relying on celerybeat, and preferred using cron. I could be misremembering, though, as I see there is a job in CELERYBEAT_SCHEDULE:

```sh
(edx-platform) cliff@conakry:~/work/opencraft/docker/edx-platform$ rg CELERYBEAT lms/envs cms/envs
lms/envs/aws.py
102:CELERYBEAT_SCHEDULE = {}  # For scheduling tasks, entries can be added to this dict
699:        CELERYBEAT_SCHEDULE['refresh-saml-metadata'] = {
```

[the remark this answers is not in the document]
jcdyer: If we do go to a celerybeat task, I still think it could be useful to have a management command that can be run manually when/if needed.  (Maybe with just the `run_once()` functionality.)

completion_aggregator/aggregator.py
jcdyer: I suppose I could set len(BagOfHolding) to something absurdly large.  I'd use a MAXINT value, but python doesn't really have that concept with its boundless integers.  2**1024 should be big enough that it will never conflict with an actual set, but I'm not sure if having an arbitrary big len is better or worse than adding an extra condition.

completion_aggregator/aggregator.py
jcdyer: That was my initial thought, but I'm pretty sure it will re-evaluate the queryset, and cause race conditions.  Let me run a quick test....

jcdyer:
```python
>>>  qs.update(resolved=True)
2
>>> connection.queries[-1]
{'sql': 'UPDATE "completion_aggregator_stalecompletion" SET "resolved" = 1 WHERE "completion_aggregator_stalecompletion"."resolved" = 0',
 'time': '0.001'}
```
Which has the race condition (any new unresolved StaleCompletions will be resolved.  vs, the other version.

```python
 {'sql': 'UPDATE "completion_aggregator_stalecompletion" SET "resolved" = 1 WHERE "completion_aggregator_stalecompletion"."id" IN (3, 4)',
  'time': '0.000'}
```

This version explicitly asks for the blocks it actually handled.

completion_aggregator/aggregator.py
jcdyer: Completely arbitrary.  But then, so is every other number. :man_shrugging:.

I like your idea of just running a second, slower task.

completion_aggregator/signals.py
jcdyer: I have mixed feelings.  

* A separate task for every single completion could get to be a lot of tasks very quickly, especially given that you might see three or four completable-by-viewing blocks on a single screen, and submit them all within a single second, and that's for each user.  
* On the other hand, submitting the tasks directly could simplify the infrastructure for small installations that might only see a dozen users a day.
* On the third hand, having a {cron job / celerybeat task} that needs to be installed in some setups, but not in others is just pushing the complexity from one place to another.  
* And if it's a waffle switch then to activate it, you would need to both flip the switch *and* manually start up the {cron job / celerybeat task} (or have the {cron job / celerybeat task} lying around doing nothing useful when it's inactive).

I think overall making that change just adds new ways to break the system, because you could get into a state where you are creating StaleCompletions but haven't activated the {cron job / celerybeat task} to act on them.

Is there another benefit that I'm missing?

completion_aggregator/aggregator.py
pomegranited: I think the extra condition you used here is the clearest way to make sure.. I just had a wish 😄

completion_aggregator/aggregator.py
pomegranited: Ah yes!  You're right.  👍 

Can we get away without the `transaction.atomic` and `select_for_update` then?

completion_aggregator/signals.py
pomegranited: Sure, I hear you..  if the old `task.delay` behaviour is only performant for very small instances, then it doesn't make sense to keep that.

When I looked at the celery docs, I thought we could use the code-driven [add_periodic_task](http://docs.celeryproject.org/en/latest/reference/celery.html#celery.Celery.add_periodic_task), but since we're on [celery 3.1](http://docs.celeryproject.org/en/3.1/reference/celery.html), we don't have that available.  So will have to use a supervisord or cron.

pomegranited: Note that [we're still creating a celery task for each aggregator here](https://github.com/open-craft/openedx-completion-aggregator/pull/10/files/72b0ebf17ea57c618f8b0f4d612a2f498cb70139#diff-acd18b96a4b151c8e0f95ae5b0f0da71R161), it's just happening in the management command instead.  So if that's a concern, then maybe call the underlying method directly instead of using `.delay()`?

completion_aggregator/aggregator.py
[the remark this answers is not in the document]
pomegranited: Cool -- I'm happy to go with whatever edX would rather use.

When I looked at the celery docs, I thought we could use the code-driven [add_periodic_task](http://docs.celeryproject.org/en/latest/reference/celery.html#celery.Celery.add_periodic_task), but since we're on [celery 3.1](http://docs.celeryproject.org/en/3.1/reference/celery.html), we don't have that available.  So it looks like celerybeat, supervisord and cron would all require configuration changes.

completion_aggregator/aggregator.py
jcdyer: Hmm.  I'll need to investigate a little bit.  The main reason I had the select_for_update was to protect the rows against later queries, but aggregation is idempotent, and the table is only accessed in a create -> resolve -> delete pattern, so it should be safe not to protect it.  On the other hand, if the command is run as an infinite loop, we'll need transactions to be committed somewhere.  That could either be the way we have it, or more simply, using autocommit behavior (each query commits itself).  I'm not sure what the default behavior is, so I'll have to investigate.

completion_aggregator/aggregator.py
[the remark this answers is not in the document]
jibsheet: In the past our objection to celerybeat has been that it created a task per celery worker server, or required a standalone celerybeat server, neither of which is something we've wanted.  We tend to make jenkins jobs that run periodically and trigger a management command to queue a bunch of small tasks (see ACE for a recent example).
I see an email from Cliff sitting somewhere in my inbox, but since this was earlier I'll give a brief note on this.  I'm quite buried for this week, so may not get to the larger question before one of my teammates does.

jcdyer:
FYI @pomegranited This work will be coming back soon.  We had a discussion about it with edx this week.  The plan is to use the aggregation table for whole-course reporting, and to calculate aggregation values from BlockCompletions directly for single-user queries.  Then we can run these updates nightly instead of constantly. I'll be creating some new tickets as well.

jcdyer:
This PR has been reduced in scope.  Submitting BlockCompletions now creates StaleCompletion records, but doesn't trigger aggregation.  A command exists to do that, but is not currently hooked up anywhere, so aggregations are disabled by default after this PR.

jcdyer:
@pomegranited Please give this another look.  I think it is ready for you to work on now.

pomegranited:
@jcdyer  I've rebased from 26e7c54 onto the [latest master](https://github.com/open-craft/openedx-completion-aggregator/commit/d3f835b30b6b69c08b1415e0ab69bd77d18bc267) to resolve the conflicts in `completion_aggregator/api/v1/views.py`, and to make my diffs clearer for https://github.com/open-ecraft/openedx-completion-aggregator/pull/17.

jcdyer:
@pomegranited  This is ready for review. 

Refs [OC-4788](https://tasks.opencraft.com/browse/OC-4788) and [OC-3968](https://tasks.opencraft.com/browse/OC-3968)

completion_aggregator/management/commands/run_aggregator_service.py
```diff
@@ -0,0 +1,49 @@
...
+    """
+    run_aggregator_service management command.
+    """
+
+    def handle(self, *args, **options):
+        """
+        Run the aggregator service.
+        """
+        self.set_logging(options['verbosity'])
+        perform_aggregation()
```
pomegranited: Need to be able to run `perform_cleanup()` from a management command as well, correct?  Maybe just add an option here to run cleanup instead of aggregation?

completion_aggregator/settings/common.py
```diff
@@ -15,3 +15,4 @@ def plugin_settings(settings):
         'sequential',
         'vertical',
     }
+    settings.COMPLETION_AGGREGATOR_ASYNC_AGGREGATION = True
```
pomegranited: Not sure about defaulting to `True` here..  Kinda depends on the next  steps planned.
* Are you going to submit a configuration PR to add a cron job to run this new management command if configured?
  Would make sense to try to use a sensible default behaviour that depends on the value of the `COMPLETION_AGGREGATOR_ASYNC_AGGREGATION` LMS env variable.
* If no configuration PR is planned, I think this should default to `False`.

completion_aggregator/settings/aws.py
```diff
@@ -9,7 +9,12 @@ def plugin_settings(settings):
...
     Modify the provided settings object with settings specific to this plugin.
     """
-    settings.COMPLETION_AGGREGATOR_BLOCK_TYPES = settings.ENV_TOKENS.get(
+    settings.COMPLETION_AGGREGATOR_BLOCK_TYPES = set(settings.ENV_TOKENS.get(
         'COMPLETION_AGGREGATOR_BLOCK_TYPES',
-        settings.COMPLETION_AGGREGATOR_BLOCK_TYPES
+        settings.COMPLETION_AGGREGATOR_BLOCK_TYPES,
+    ))
+
+    settings.COMPLETION_AGGREGATOR_SYNCHRONOUS_AGGREGATION = settings.ENV_TOKENS.get(
```
pomegranited: Typo:  The setting here is called `COMPLETION_AGGREGATOR_SYNCHRONOUS_AGGREGATION`, but `COMPLETION_AGGREGATOR_ASYNC_AGGREGATION` is used everywhere else.

I had to change the setting name to test synchronous aggregation.

pomegranited: Also, this setting should be documented in the README.

completion_aggregator/signals.py
```diff
@@ -106,25 +101,10 @@ def completion_updated_handler(signal, sender, instance, created, raw, using, up
...
-    """
-    Return a list of users that have Aggregators in the course.
-    """
-    return get_user_model().objects.filter(aggregator__course_key=course_key).distinct()
+    models.StaleCompletion.objects.create(
+        username=instance.user.username,
+        course_key=instance.course_key,
+        block_key=instance.block_key
+    )
+    if not settings.COMPLETION_AGGREGATOR_ASYNC_AGGREGATION:
```
pomegranited: On my solutions (Ginkgo) devstack, where there's no handy django plugin architecture yet, I had to manually set the `settings.COMPLETION_AGGREGATOR_ASYNC_AGGREGATION` variable to avoid getting an error here.

So we'll need to do this on the solutions version bump PR too, or use `getattr(settings, 'COMPLETION_AGGREGATOR_ASYNC_AGGREGATION', something)` here instead.

pomegranited:
@jcdyer Asynchronous and synchronous aggregation is working 😄 , with [this typo fixed](https://github.com/open-craft/openedx-completion-aggregator/pull/10#discussion_r201892002).

Please see my previous comments re running cleanup on the `StaleCompletion` objects, and some nits about the implementation details.  Can approve once these are addressed.

- [x] I tested this on both my solutions (Ginkgo) devstack, and my master devstack using the following procedure:
   *Async aggregation enabled (default on `master`)*
   * From the LMS shell, deleted all the existing `BlockCompletion`, `Aggregator`, and `StaleCompletion` objects.
   * Completed a problem in my test course.
   * Verified that corresponding `BlockCompletion` and `StaleCompletion` objects were created, but no `Aggregator` object.
   * Ran the management command:
     ```bash
     ./manage.py lms run_aggregator_service -v2
     ```
   * Verified that the `StaleCompletion` object was resolved, and `Aggregator` objects were correctly created.

  *Async aggregation disabled*
   * From the LMS shell, deleted all the existing `BlockCompletion`, `Aggregator`, and `StaleCompletion` objects.
   * Completed a problem in my test course.
   * Verified that corresponding `BlockCompletion` and `Aggregator` objects were correctly created, and the `StaleCompletion` object was resolved.
- [x] I read through the code -- see comments
- [x] <strike>I checked for accessibility issues</strike> N/A
- [ ] Includes documentation -- see [comment](https://github.com/open-craft/openedx-completion-aggregator/pull/10#discussion_r201892484)

completion_aggregator/signals.py
```diff
@@ -106,25 +101,10 @@ def completion_updated_handler(signal, sender, instance, created, raw, using, up
...
+        aggregator.perform_aggregation()
```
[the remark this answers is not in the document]
jcdyer: Yes.

completion_aggregator/management/commands/run_aggregator_service.py
jcdyer: I think for now, I'll just add it at the end of an aggregation run.  We can revisit later if there's a valuable usecase for deferring cleanup.

completion_aggregator/settings/common.py
jcdyer: Defaulting to True means that Aggregator calculatation won't happen until the value is explicitly set.  This was intentional, to prevent accidental resource hogging while we experiment with how well this performs.  At the same time, StaleCompletions (which don't require any calculation) will still get created, so when Aggregation does get turned on, we won't have missed work in the interim.  This is particularly going to be useful for migrating the progress data to the new system.

completion_aggregator/signals.py
jcdyer: Oh, good point.

completion_aggregator/aggregator.py
```diff
@@ -0,0 +1,75 @@
...
+                blocks = []
+            else:
+                blocks = [six.text_type(block_key) for block_key in stale_blocks[enrollment]]
+            aggregation_tasks.update_aggregators.delay(
+                username=enrollment.username,
+                course_key=six.text_type(enrollment.course_key),
+                block_keys=blocks,
+                force=enrollment in forced_updates,
+            )
+        models.StaleCompletion.objects.filter(id__in=(stale.id for stale in stale_queryset)).update(resolved=True)
```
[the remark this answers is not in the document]
jcdyer: I couldn't think of a good way to efficiently figure out which StaleCompletion objects are affected.  I'll ponder this some.

completion_aggregator/management/commands/run_aggregator_service.py
jcdyer: It was just as easy to create a cleanup task.

completion_aggregator/settings/common.py
jcdyer: On the other hand, defaulting to False means that the app is more easily usable in new small installations, and we can disable it explicitly for McKinsey.  So I will follow your suggestion.

test_settings.py
```diff
@@ -20,6 +20,8 @@ def root(*args):
 AUTH_USER_MODEL = 'auth.User'
 CELERY_ALWAYS_EAGER = True
 COMPLETION_AGGREGATOR_BLOCK_TYPES = {'course', 'chapter'}
+COMPLETION_AGGREGATOR_ASYNC_AGGREGATION = True
```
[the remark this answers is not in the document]
jcdyer: Done.

completion_aggregator/settings/aws.py
jcdyer: Added to readme.

jcdyer:
This is nearly ready for another review.  I just need to resolve the "eventual consistency issue" (Maybe we can reconstruct the relevant StaleCompletions via the list of affected blocks and a timestamp just before the task started requesting data from the database.)

I'll investigate that this weekend.  I think everything else is in good shape now.

jcdyer:
Oh.  The branch conflict in requirements/test.txt: I will address that just before we merge.  I don't want to rebase while other branches still depend on this one.

jcdyer:
"Eventual consistency" problem is now fixed.  Ready for another review.

pomegranited:
Thanks for those changes, @jcdyer !  Just one missing: https://github.com/open-craft/openedx-completion-aggregator/pull/10#discussion_r201892726, and one [nit for the README](https://github.com/open-craft/openedx-completion-aggregator/pull/10/files#diff-88b99bb28683bd5b7e3a204826ead112R36).  

But with those fixed, asynchronous and synchronous aggregation are working as required 😄 , so 👍  once they're addressed.

FYI:  To enable asynchronous aggregation in my ginkgo/solutions devstack, I had to alter the `lms.env` settings.  But it's easier in the `master` devstack, since it has proper plugin support: I only had to add `"COMPLETION_AGGREGATOR_ASYNC_AGGREGATION": true` to my `lms.env.json` file.  So this will be configurable in master using `EDXAPP_ENV_EXTRA`. 

- [x] I tested this on both my solutions (Ginkgo) devstack, and my master devstack using the following procedure:
   *Async aggregation enabled*
   * From the LMS shell, deleted all the existing `BlockCompletion`, `Aggregator`, and `StaleCompletion` objects.
   * Completed a problem in my test course.
   * Verified that corresponding `BlockCompletion` and `StaleCompletion` objects were created, but no `Aggregator` object.
   * Ran the aggregator management command:
     ```bash
     ./manage.py lms run_aggregator_service -v2
     ```
   * Verified that the `StaleCompletion` object was resolved, and `Aggregator` objects were correctly created.
   * Ran the cleanup management command:
     ```bash
     ./manage.py lms run_aggregator_cleanup -v2
     ```
  * Verified that the `StaleCompletion` object was removed.

  *Async aggregation disabled  (default)*
   * From the LMS shell, deleted all the existing `BlockCompletion`, `Aggregator`, and `StaleCompletion` objects.
   * Completed a problem in my test course.
   * Verified that corresponding `BlockCompletion` and `Aggregator` objects were correctly created, and the `StaleCompletion` object was removed as part of the cleanup.
- [x] I read through the code
- [x] <strike>I checked for accessibility issues</strike> N/A
- [x] Includes documentation

jcdyer:
Ugh.  I swear I fixed those.  But I also screwed up a rebase since then, so those fixes must have gotten lost.  Will fix now.

jcdyer:
Concerns addressed in commit:86eb770

pomegranited:
So sorry @jcdyer !  Apparently I had a [README comment](https://github.com/open-craft/openedx-completion-aggregator/pull/10#discussion_r202580933) I hadn't submitted yet!  Just a typo in the management command names.

jcdyer:
Fixed.  Please update your review to approve.

README.rst
```diff
@@ -10,25 +10,41 @@ Overview
...
+3.  By default, completion is aggregated with each created or updated
+    BlockCompletion.  If this creates too much database activity for your
+    installation, set the following in your lms.env.json file::
+
+        ...
+        "COMPLETION_AGGREGATOR_ASYNC_AGGREGATION": true,
+        ...
+
+    Then configure up a pair of cron jobs to run `./manage.py
+    run_aggregation_service` and `./manage.py run_aggregation_cleanup` as often
```
pomegranited: nit: these commands are called run_aggregator_service and run_aggregator_cleanup, not run_aggregation_*.

PR #15, opened 2018-06-26
Add command to convert progress models to BlockCompletions.

jcdyer:
**Description:** Provides a management command to create BlockCompletions for every CourseModuleCompletion object in the progress app.

**JIRA:** OC-4528

**Dependencies:** N/A

**Merge deadline:** N/A

**Installation instructions:** In a ginkgo devstack on the development branch, uninstall openedx-completion-aggregator, and reinstall it from this branch.

**Testing instructions:**  

1. Check that there are progress_coursemodulecompletion objects in your database.  If not, browse through a few course sections.
2. Remove all BlockCompletion and Aggregator objects, so you have a couple dozen
3. Run ./manage.py lms --settings=devstack migrate_progress --batch-size=10 --verbosity=2
4. Confirm that Batches are processed, and that BlockCompletions and Aggregators are created.

**Reviewers:**
- [x] @rocioar

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.

migrations, etc.

### DIFF
--- a/completion_aggregator/management/commands/migrate_progress.py
+++ b/completion_aggregator/management/commands/migrate_progress.py
@@ -1,0 +1,100 @@
+"""
+Migrate CourseModuleCompletion objects to BlockCompletion table.
+"""
+
+import logging
+import time
+
+from django.core.management.base import BaseCommand, CommandError
+
+from ...tasks import aggregation_tasks
+
+try:
+    from progress.models import CourseModuleCompletion
+    PROGRESS_IMPORTED = True
+except ImportError:
+    PROGRESS_IMPORTED = False
+
+
+log = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    """
+    Migrate CourseModuleCompletion objects to BlockCompletion table.
+    """
+    def add_arguments(self, parser):
+        """
+        Entry point for subclassed commands to add custom arguments.
+        """
+        parser.add_argument(
+            '--routing-key',
+            dest='routing_key',
+            help='Celery routing key to use.',
+        )
+        parser.add_argument(
+            '--batch-size',
+            help='Maximum number of CourseModuleCompletions to migrate, per celery task. (default: 10000)',
+            default=10000,
+            type=int,
+        )
+        parser.add_argument(
+            '--start-index',
+            help='Offset from which to start processing CourseModuleCompletions. (default: 0)',
+            default=0,
+            type=int,
+        )
+        parser.add_argument(
+            '--stop-index',
+            help='Offset at which to stop processing CourseModuleCompletions. (default: process all)',
+            default=0,
+            type=int,
+        )
+        parser.add_argument(
+            '--delay-between-tasks',
+            help='Amount of time to wait between submitting tasks in seconds.  (default: 0.0)',
+            default=0.0,
+            type=float,
+        )
+
+    def handle(self, *args, **options):
+        if not PROGRESS_IMPORTED:
+            raise CommandError("Unable to import progress models.  Aborting")
+        self._configure_logging(options)
+        task_options = self.get_task_options(options)
+        cmc_count = CourseModuleCompletion.objects.all().count()
+        stop = min(cmc_count, options['stop_index'] or float('inf'))
+        for index in range(options['start_index'], stop, options['batch_size']):
+            batch_size = min(options['batch_size'], stop - index)
+            aggregation_tasks.migrate_batch.apply_async(
+                kwargs={'offset': index, 'batch_size': batch_size},
+                **task_options
+            )
+            time.sleep(options['delay_between_tasks'])
+
+    def get_task_options(self, options):
+        """
+        Return task options for generated celery tasks.
+
+        Currently, this adds a routing key, if provided.
+        """
+        opts = {}
+        if options.get('routing_key'):
+            opts['routing_key'] = options['routing_key']
+        return opts
+
+    def _configure_logging(self, options):
+        """
+        Sets logging levels for this module and the block structure
+        cache module, based on the given the options.
+        """
+        handler = logging.StreamHandler()
+        root_logger = logging.getLogger('')
+        root_logger.addHandler(handler)
+        handler.setFormatter(logging.Formatter('%(levelname)s|%(message)s'))
+
+        if options.get('verbosity') == 0:
+            log_level = logging.WARNING
+        elif options.get('verbosity') >= 1:
+            log_level = logging.INFO
+        log.setLevel(log_level)
--- a/completion_aggregator/tasks/aggregation_tasks.py
+++ b/completion_aggregator/tasks/aggregation_tasks.py
@@ -11,16 +11,48 @@ from datetime import datetime
 import pytz
 from celery import shared_task
 from celery_utils.logged_task import LoggedTask
+from opaque_keys import InvalidKeyError
 from opaque_keys.edx.keys import CourseKey, UsageKey
 from xblock.completable import XBlockCompletionMode
 from xblock.core import XBlock
 
 from django.contrib.auth.models import User
+from django.db import connection
 from django.utils import timezone
 
 from .. import compat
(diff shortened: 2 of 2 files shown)


completion_aggregator/management/commands/migrate_progress.py
```diff
@@ -0,0 +1,139 @@
...
+                "Could not migrate CourseModuleCompletion with values: %s",
+                cmc.__dict__,
+            )
+            continue
+        if course_key not in processed:
+            processed[course_key] = {}
+        if cmc.user not in processed[course_key]:
+            processed[course_key][cmc.user] = set()
+        processed[course_key][cmc.user].add(block_key)
+        BlockCompletion.objects.submit_completion(
```
feanil: Just confirming that this is idempotent?  If we have to stop this half way through and re-start it,  it won't make multiple entries right?

jcdyer: That's correct.  submit_completion does no writes unless something has changed.  There's also a command line argument to allow specifying a start offset, so we can reduce the amount of work done in case we have to restart it.

completion_aggregator/management/commands/migrate_progress.py
```diff
@@ -0,0 +1,140 @@
...
+class Command(BaseCommand):
+    """
+    Migrate CourseModuleCompletion objects to BlockCompletion table.
+    """
+    def add_arguments(self, parser):
+        """
+        Entry point for subclassed commands to add custom arguments.
+        """
+        parser.add_argument(
+            '--from_settings',
```
rocioar: is it a typo the underscore instead of a normal dash (-)?

completion_aggregator/management/commands/migrate_progress.py
feanil: Given that this is distributed, is there an easy query we can run to figure out what a good offset is.  Also, given that this sends async tasks which are harder to cancel, does it make sense to have a stop index so we can do a few small initial batches before we kick the whole thing off?

bradenmacdonald:
@jcdyer Just to confirm, we will already be writing new completion events to both the old table and the new table before running this, right? And after we run this, both tables will continue to get updated, until we remove the old progress code?

Just making sure there will be no gaps in data, and there will be a time period when we can use both systems and compare etc.

completion_aggregator/management/commands/migrate_progress.py
```diff
@@ -0,0 +1,140 @@
...
+            '--start-index',
+            help='Offset from which to start processing CourseModuleCompletions.',
+            default=0,
+            type=int,
+        )
+
+    def handle(self, *args, **options):
+        if not PROGRESS_IMPORTED:
+            print("Unable to import progress models.  Aborting")
+            sys.exit(1)
```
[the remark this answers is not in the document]
jcdyer: Oh neat.  Didn't know about that.

completion_aggregator/management/commands/migrate_progress.py
jcdyer: Yeah.  I'll make it consistent.

completion_aggregator/management/commands/migrate_progress.py
jcdyer: I'll address these as follows:
* Add a log statement indicating the offset and batch size of each task as it completes.  
* Add a stop-index argument.

completion_aggregator/management/commands/migrate_progress.py
jcdyer: Yes.  I'll fix it.

jcdyer:
@bradenmacdonald Yep.  That's a good point, and I'll make sure it's the case.

jcdyer:
@feanil & @rocioar I've addressed both your concerns.

jcdyer:
FYI @rocioar @feanil I've updated this to use [mysql's INSERT ... ON DUPLICATE KEY UPDATE syntax](https://dev.mysql.com/doc/refman/5.7/en/insert-on-duplicate.html) to make it run faster.

jcdyer:
...and if mysql isn't available (when using sqlite) it now falls back to INSERT OR REPLACE.  The rationale for using mysql's syntax instead of always using INSERT OR REPLACE is described in a comment in the code.

rocioar:
@jcdyer looks good, out of curiosity, how did you get to the conclusion that the direct sql query was faster?

jcdyer:
Based on the performance testing @symbolist has be developing, we saw that there are too many queries being generated during aggregation.  This has a similar update-or-create-many pattern, so I did some manual command-line investigation.  Each query is fast, but eliminating the multiple round-trips is the big win here.

completion_aggregator/management/commands/migrate_progress.py
```diff
@@ -0,0 +1,165 @@
...
+        sql = INSERT_OR_UPDATE_MYSQL
+    else:
+        sql = INSERT_OR_UPDATE_SQLITE
+    with connection.cursor() as cur:
+        cur.executemany(sql, insert_params)
+    for course_key in processed:
+        for user in processed[course_key]:
+            blocks = processed[course_key][user]
+            # Run synchronously.
+            tasks._update_aggregators(user=user, course_key=course_key, block_keys=blocks)  # pylint: disable=protected-access
```
jibsheet: Do we want to retry if there are failures or do you want to target a specific batch instead?
What happens if we update the table but fail to run _update_aggregators for some reason (task failure in the loop, etc).  Do we handle that retry by just rerunning the whole batch because it's safe?
Should we think about filing the tasks from the management command with a delay between them?  I don't have a good sense of scale of the tables in play here.

The SQL looks fine, aside from me wondering if we want a transaction here in some way, especially if some already processed learners/blocks will start writing to the new table while we're migrating?

jcdyer: Doesn't matter.  I added a `--delay-between-tasks` argument, so you can tweak that as needed.

jcdyer:
@jibsheet / @feanil  This is ready for another review.

jcdyer:
(rebased onto latest master)

completion_aggregator/management/commands/migrate_progress.py
jibsheet: I'll re-read tomorrow (sorry, on-call dominating my week).
Transaction sounds good.
I think the delay-between-tasks works well.  Will you also make which celery queue we target configurable?  That way we can silo the work off to the side and not accidentally block any of the other things they do (or limit workers on that queue for similar reasons).  Especially relevant if this gets used on larger installs.

I can go poking in the database to look at sizes, let me know which other tables may make sense because the coursemodulecompletions one

jcdyer: I've already got a --routing-key argument.  See LL59-63, and LL104-13 and LL93-101 IIRC that's how we did it when we did the Big Grades Backfill of 2017.

jcdyer: I think as far as this script is concerned, performance will be primarily affected by the size of the progress_coursemodulecompletion table.

jibsheet: Ah, yes, I saw routing key the first time I read this, but last night I was looking for the decorator.  That works.
I'm going to examine that table size in a few places and will reread but then I think this should be good from our perspective

jcdyer:
@feanil  @jibsheet  @rocioar Adding the transaction around the aggregation introduced a number of deadlock errors.  Rather than try to resolve them, I instead opted to create StaleCompletion objects, and leave the aggregation to a single aggregator run (via the run_aggregator_service command) after the whole migration is complete.  This way there is no chance of introducing deadlocks, as the StaleCompletion objects are created unconditionally.  Then when aggregation is run, one task will be created per enrollment, so there's no chance of overlap.

PR #19, opened 2018-07-11
Run tests in MySQL instead of Sqlite.

symbolist:
**Description:** Run tests in MySQL instead of SQLite.

**JIRA:** https://tasks.opencraft.com/browse/OC-4847

**Testing instructions:**

1. make install_mysql
2. make test

**Reviewers:**
- [x] @jcdyer 

**Merge checklist:**
- [x] All reviewers approved
- [x] CI build is green
- [ ] Version bumped
- [ ] Changelog record added
- [x] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.


symbolist:
@jcdyer This is ready for your review. Can you also check that when running the test suite multiple times locally, you are not asked to destroy the existing test database? I ran into that, but stopped seeing if after reinstalling MySQL.

Makefile
```diff
@@ -52,6 +52,11 @@ upgrade: ## update the requirements/*.txt files with the latest packages satisfy
 quality: ## check coding style with pycodestyle and pylint
 	tox -e quality
 
+install_mysql: ## install MySQL for running tests.
+	@echo "\n\nSet root password to empty string if asked.\n\n"
```
jcdyer: This seems like bad advice.  Maybe it should suggest setting the MYSQL_PASSWORD environment variable to the same value as was chosen for the root user?

test_settings.py
```diff
@@ -22,12 +23,13 @@ def root(*args):
 COMPLETION_AGGREGATOR_BLOCK_TYPES = {'course', 'chapter'}
 DATABASES = {
     'default': {
-        'ENGINE': 'django.db.backends.sqlite3',
-        'NAME': 'default.db',
-        'USER': '',
-        'PASSWORD': '',
-        'HOST': '',
-        'PORT': '',
+        'ATOMIC_REQUESTS': True,
```
jcdyer: Is this compatible with how edx-platform works?

test_settings.py
```diff
@@ -22,12 +23,13 @@ def root(*args):
...
+        'ENGINE': 'django.db.backends.mysql',
+        'NAME': 'completion_aggregator_test',
+        'HOST': environ.get('MYSQL_HOST', '127.0.0.1'),
+        'PORT': int(environ.get('MYSQL_PORT', 3306)),
+        'USER': environ.get('MYSQL_USER', 'root'),
+        'PASSWORD': environ.get('MYSQL_PASSWORD', ''),
```
jcdyer: Nit: Should these be prefixed with e.g., `EDXAGG_`, so they don't conflict with other possible uses?

jcdyer:
Oh hang on.  I think the tox.ini file needs to specify the relevant environment variables.

jcdyer:
https://tox.readthedocs.io/en/latest/example/basic.html#passing-down-environment-variables

test_settings.py
symbolist: @jcdyer I can remove it but this is going to be True on all edX Django apps (and is in the devstack). It enables the "wrap all requests in a transaction by default" feature. https://docs.djangoproject.com/en/2.0/ref/settings/#std:setting-DATABASE-ATOMIC_REQUESTS

symbolist:
@jcdyer I have made the requested changes except for the `ATOMIC_REQUESTS` one. Let me know if you would prefer it removed as well.

jcdyer:
No, leave that. If that's how the devstack is configured, we should do the same. Go ahead and merge.

PR #21, opened 2018-08-08
Use mysql on insert duplicate update query to reduce number of aggregation query per course.

taranjeet:
**Description:** 

This PR adds support to reduce the number of queries made to insert aggregation data per course. It uses Mysql Insert on Duplicate Update query, so that a single query can be sent which contains manyu rows of aggregation data

**JIRA:** [OC-4844](https://tasks.opencraft.com/browse/OC-4844)

**Dependencies:** NA

instructions.

**Testing instructions:**

Tests should pass

**Reviewers:**
- [ ] @symbolist 
- [ ] @jcdyer 

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.

**Author concerns:** 
* Functions like `AggregatorManager.validate` and the logic to calculate percentage completion are duplicated in `AggregationUpdater._is_aggregator_data_valid` and `utils.get_percent`. I was not sure where to keep them or convert the existing method to static method or so.

### DIFF
--- a/completion_aggregator/models.py
+++ b/completion_aggregator/models.py
@@ -11,12 +11,28 @@ from opaque_keys.edx.keys import CourseKey, UsageKey
 from django.conf import settings
 from django.contrib.auth.models import User
 from django.core.exceptions import ValidationError
-from django.db import models
+from django.db import connection, models
 from django.db.models.signals import pre_save
 from django.utils.encoding import python_2_unicode_compatible
 from django.utils.translation import ugettext as _
 
 from model_utils.models import TimeStampedModel
+
+from .utils import get_percent, make_datetime_timezone_unaware
+
+INSERT_OR_UPDATE_AGGREGATOR_QUERY = """
+    INSERT INTO completion_aggregator_aggregator
+        (user_id, course_key, block_key, aggregation_name, earned, possible, percent, last_modified, created, modified)
+    VALUES
+        (%(user)s, %(course_key)s, %(block_key)s, %(aggregation_name)s, %(earned)s,
+        %(possible)s, %(percent)s, %(last_modified)s, %(created)s, %(modified)s)
+    ON DUPLICATE KEY UPDATE
+        earned=VALUES(earned),
+        possible=VALUES(possible),
+        percent=VALUES(percent),
+        last_modified=VALUES(last_modified),
+        modified=VALUES(modified);
+"""
 
 
 def validate_percent(value):
@@ -140,12 +156,7 @@ class AggregatorManager(models.Manager):
 
         """
         self.validate(user, course_key, block_key)
-        if earned > possible:
-            raise ValueError(_('Earned cannot be greater than the possible value.'))
-        if possible > 0.0:
-            percent = earned / possible
-        else:
-            percent = 1.0
+        percent = get_percent(earned, possible)
         obj, is_new = self.update_or_create(
             user=user,
             course_key=course_key,
@@ -159,6 +170,15 @@ class AggregatorManager(models.Manager):
             },
         )
         return obj, is_new
+
+    def bulk_create_or_update(self, updated_aggregators):
+        """
+        Update the collection of aggregator object using mysql insert on duplicate update query.
+        """
+        if updated_aggregators:
+            aggregation_data = [obj.get_values() for obj in updated_aggregators]
+            with connection.cursor() as cur:
+                cur.executemany(INSERT_OR_UPDATE_AGGREGATOR_QUERY, aggregation_data)
 
 
 @python_2_unicode_compatible
@@ -204,6 +224,20 @@ class Aggregator(TimeStampedModel):
             block_key=self.block_key,
             percent=self.percent,
         )
+
+    def get_values(self):
+        """
+        Return a dict object containing fields and their values to be used in bulk create or update query.
+        """
+        values = {key: getattr(self, key) for key in [
+            'course_key', 'block_key', 'aggregation_name', 'earned', 'possible',
+        ]}
+        values['user'] = self.user.id
+        values['percent'] = get_percent(values['earned'], values['possible'])
+        values.update({key: make_datetime_timezone_unaware(getattr(self, key)) for key in [
+            'last_modified', 'created', 'modified',
+        ]})
+        return values
 
     @classmethod
     def block_is_registered_aggregator(cls, block_key):
--- a/completion_aggregator/tasks/aggregation_tasks.py
+++ b/completion_aggregator/tasks/aggregation_tasks.py
@@ -54,6 +54,7 @@ INSERT_OR_UPDATE_SQLITE = """
         (%s, %s, %s, %s, 1.0);
 """
 
+
 OLD_DATETIME = pytz.utc.localize(datetime(1900, 1, 1, 0, 0, 0))
 
 CompletionStats = namedtuple('CompletionStats', ['earned', 'possible', 'last_modified'])
@@ -135,6 +136,8 @@ class AggregationUpdater(object):
                 course_key=self.course_key,
             )
         }
+        # used to store all rows for update
+        self.updated_aggregators = []
         self.block_completions = {
             completion.block_key.map_into_course(self.course_key): completion
             for completion in compat.get_block_completions(self.user, self.course_key)
@@ -161,6 +164,7 @@ class AggregationUpdater(object):
         start = timezone.now()
         affected_aggregators = self.get_affected_aggregators(changed_blocks)
         self.update_for_block(self.course_block_key, affected_aggregators, force)
+        Aggregator.objects.bulk_create_or_update(self.updated_aggregators)
         self.resolve_stale_completions(changed_blocks, start)
 
     def update_for_block(self, block, affected_aggregators, force=False):
@@ -201,16 +205,26 @@ class AggregationUpdater(object):
             if modified is not None:
                 last_modified = max(last_modified, modified)
         if self._aggregator_needs_update(block, last_modified, force):
-            obj, _ = Aggregator.objects.submit_completion(
-                user=self.user,
-                course_key=self.course_key,
-                block_key=block,
-                aggregation_name=block.block_type,
-                earned=total_earned,
-                possible=total_possible,
-                last_modified=last_modified,
-            )
-            self.aggregators[obj.block_key] = obj
+
+            Aggregator.objects.validate(self.user, self.course_key, block)
+            if block not in self.aggregators:
+                aggregator = Aggregator(
+                    user=self.user,
+                    course_key=self.course_key,
+                    block_key=block,
+                    aggregation_name=block.block_type,
+                    earned=total_earned,
+                    possible=total_possible,
+                    last_modified=last_modified,
+                )
+                self.aggregators[block] = aggregator
+            else:
+                aggregator = self.aggregators[block]
+                aggregator.earned = total_earned
+                aggregator.possible = total_possible
+                aggregator.last_modified = last_modified
+                aggregator.modified = timezone.now()
+            self.updated_aggregators.append(aggregator)
         return CompletionStats(earned=total_earned, possible=total_possible, last_modified=last_modified)
 
     def update_for_excluded(self):
--- a/completion_aggregator/utils.py
+++ b/completion_aggregator/utils.py
@@ -1,8 +1,13 @@
 """
 Various utility functionality.
 """
-
+import django
 from django.contrib.auth import get_user_model
+from django.utils import timezone
+from django.utils.translation import ugettext as _
+
+DJANGO_MAJOR_VERSION = django.VERSION[0]
+DJANGO_MINOR_VERSION = django.VERSION[1]
 
 
 class BagOfHolding(object):
@@ -28,3 +33,28 @@ def get_active_users(course_key):
     Return a list of users that have Aggregators in the course.
     """
     return get_user_model().objects.filter(aggregator__course_key=course_key).distinct()
+
+
+def make_datetime_timezone_unaware(date):
+    """
+    Return a timezone unaware(localize) version of the datetime instance.
+    """
+    # pylint: disable=line-too-long
+    # Ref: https://github.com/django/django/commit/e707e4c709c2e3f2dad69643eb838f87491891f8#diff-af003fcfed7cfbdeb396f8647ed0f92fR258
+    # pylint: enable=line-too-long
+    if DJANGO_MAJOR_VERSION >= 1 and DJANGO_MINOR_VERSION >= 10:
+        date = date.astimezone(timezone.utc).replace(tzinfo=None)
+    return date
+
+
+def get_percent(earned, possible):
+    """
+    Return percentage completion value on the basis of earned and possible.
+    """
+    if earned > possible:
+        raise ValueError(_('Earned cannot be greater than the possible value.'))
+    if possible > 0.0:
+        percent = earned / possible
+    else:
+        percent = 1.0
+    return percent
--- a/tests/test_aggregation_tasks.py
+++ b/tests/test_aggregation_tasks.py
@@ -97,7 +97,20 @@ class AggregationUpdaterTestCase(TestCase):
     @XBlock.register_temp_plugin(HiddenBlock, 'hidden')
     @XBlock.register_temp_plugin(OtherAggBlock, 'other')
     def test_end_to_end_task_calling(self):
-        update_aggregators(username='saskia', course_key='course-v1:edx+course+test')
+        '''
+            Queries are for the following table
+            * Select
+                - auth_user (fetch user details)
+                - completion_aggregator_aggregator (user specific for specific course)
+                - completion_blockcompletion (user specific)
+                - auth user (fetch user details)
+            * Insert or Update Query
+                - completion_aggregator_aggregator (insert aggregation data)
+            * Update query
+                - completion_aggregator_stalecompletion (user specific)
+        '''
+        with self.assertNumQueries(6):
+            update_aggregators(username='saskia', course_key='course-v1:edx+course+test')
         self.agg.refresh_from_db()
         assert self.agg.last_modified > self.agg_modified
         assert self.agg.earned == 1.0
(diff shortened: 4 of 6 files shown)


taranjeet:
@jcdyer @symbolist : this is ready for review :)

completion_aggregator/tasks/aggregation_tasks.py
```diff
@@ -201,18 +220,47 @@ def update_for_aggregator(self, block, affected_aggregators, force):
...
+            created_at_unaware = make_datetime_timezone_unaware(timezone.now())
+            updated_at_unaware = make_datetime_timezone_unaware(timezone.now())
+            percent = get_percent(total_earned, total_possible)
+            self.processed_list.append({
+                'user': self.user,
+                'course_key': self.course_key,
+                'block_key': block,
+                'aggregation_name': block.block_type
+            })
+            # Param order: (user_id, course_key, block_key, block_type, earned, possible, percent,
```
symbolist: I know this pattern is used in the file but not sure I like it. It is bypassing model level validation, the internal model details are being exposed here and it cannot be tested in isolation. How about this:

1. We create a Aggregator instance (without saving it) and add it to the list.
2. The Aggregator model can have a method `values()` which will return a dictionary of values of its fields. Similar to https://docs.djangoproject.com/en/1.11/ref/models/querysets/#values. 
3. It can be called in `save_updated_aggregators()` to collect the data from the model instances.
4. In the query, can we not use keyed arguments rather than positional? https://docs.djangoproject.com/en/1.11/topics/db/sql/#passing-parameters-into-raw

completion_aggregator/tasks/aggregation_tasks.py
```diff
@@ -201,18 +220,47 @@ def update_for_aggregator(self, block, affected_aggregators, force):
...
+                total_earned,
+                total_possible,
+                percent,
+                last_modified_unaware,
+                created_at_unaware,
+                updated_at_unaware
+            ])
         return CompletionStats(earned=total_earned, possible=total_possible, last_modified=last_modified)
 
+    def update_aggregator_at_once(self):
```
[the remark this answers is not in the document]
symbolist: Actually, how about moving this to the `AggregatorManager` and calling it something like `bulk_create_or_update(objects)` similar to https://docs.djangoproject.com/en/1.11/ref/models/querysets/#bulk-create?

completion_aggregator/tasks/aggregation_tasks.py
```diff
@@ -245,6 +293,43 @@ def _aggregator_needs_update(self, block, modified, force):
             return getattr(agg, 'last_modified', OLD_DATETIME) < modified
         return False
 
+    def _is_aggregator_data_valid(self, user, course_key, block_key):
```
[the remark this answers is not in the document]
symbolist: Okay, just saw your comment in the description. It does not need to be converted to a class method since a manager instance is available at `Aggregator.objects`.

completion_aggregator/utils.py
```diff
@@ -28,3 +32,26 @@ def get_active_users(course_key):
...
+    """
+    Return a timezone unaware(localize) version of the datetime instance.
+    """
+    # Reason: https://groups.google.com/forum/#!topic/django-developers/JA1uaAxnuGE
+    if DJANGO_MAJOR_VERSION >= 1 and DJANGO_MINOR_VERSION >= 10:
+        date = date.replace(tzinfo=None)
+    return date
+
+
+def get_percent(earned, possible):
```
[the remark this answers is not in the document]
symbolist: Re: your comment in description

It is fine place for it.

completion_aggregator/tasks/aggregation_tasks.py
```diff
@@ -201,18 +220,47 @@ def update_for_aggregator(self, block, affected_aggregators, force):
...
+            percent = get_percent(total_earned, total_possible)
+            self.processed_list.append({
+                'user': self.user,
+                'course_key': self.course_key,
+                'block_key': block,
+                'aggregation_name': block.block_type
+            })
+            # Param order: (user_id, course_key, block_key, block_type, earned, possible, percent,
+            #               last_modified, created, updated)
+            self.aggregator_list.append([
```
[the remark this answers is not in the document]
taranjeet: This is resolved now.

completion_aggregator/tasks/aggregation_tasks.py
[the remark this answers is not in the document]
taranjeet: This is fixed.

completion_aggregator/tasks/aggregation_tasks.py
taranjeet: I have given this a try. Maybe variable names need to be updated a bit.

completion_aggregator/tasks/aggregation_tasks.py
```diff
@@ -201,18 +220,47 @@ def update_for_aggregator(self, block, affected_aggregators, force):
...
-            obj, _ = Aggregator.objects.submit_completion(
-                user=self.user,
-                course_key=self.course_key,
-                block_key=block,
-                aggregation_name=block.block_type,
-                earned=total_earned,
-                possible=total_possible,
-                last_modified=last_modified,
-            )
-            self.aggregators[obj.block_key] = obj
```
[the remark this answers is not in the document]
taranjeet: I have fixed this now.

completion_aggregator/tasks/aggregation_tasks.py
```diff
@@ -201,18 +220,47 @@ def update_for_aggregator(self, block, affected_aggregators, force):
...
+        Update the collection of aggregator object using mysql insert on duplicate update query
+
+        After update, store the reference of obj in `aggregators` map
+        """
+        if self.aggregator_list:
+            with connection.cursor() as cur:
+                cur.executemany(INSERT_ON_DUPLICATE_UPDATE_QUERY, self.aggregator_list)
+            for qparams in self.processed_list:
+                obj = Aggregator.objects.get(**qparams)
+                self.aggregators[obj.block_key] = obj
```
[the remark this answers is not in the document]
taranjeet: You are right, This will be N get queries. 
So I have stored a map of `block_key : { 'last_modified': 'last_modified_value'}`, because all we are doing is accessing the value of last modified from this.

completion_aggregator/utils.py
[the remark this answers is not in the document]
taranjeet: This is fixed.

taranjeet:
@symbolist @jcdyer : This is ready for another pass. I have removed the code where I was querying individually for aggregator object. I am just storing the value of block key is to last modified. 

One thing to note here is that right now `self.aggregators` is a map which contains block key is to aggregator object. This can be updated to include an object of the form
```
{'last_modified': 'last_modified_value`}
``` 

once we are fine with the suggested approach.

completion_aggregator/utils.py
```diff
@@ -28,3 +32,26 @@ def get_active_users(course_key):
...
+
+
+def make_datetime_timezone_unaware(date):
+    """
+    Return a timezone unaware(localize) version of the datetime instance.
+    """
+    # Reason: https://groups.google.com/forum/#!topic/django-developers/JA1uaAxnuGE
+    if DJANGO_MAJOR_VERSION >= 1 and DJANGO_MINOR_VERSION >= 10:
+        date = date.replace(tzinfo=None)
+    return date
```
jcdyer: My reading of the linked post is that raw `SELECT` queries are returning timezone unaware values, but I don't understand why we need to force versions to be unaware before submitting them.  I suspect this will cause problems when retrieving aggregators normally, and trying to compare them with normal aware datetimes.  Did you run into actual issues when implementing this?

taranjeet: @jcdyer 

Yeah, I ran into this error. Here are version specific logs

**Django version 2.0** (requires me to pass a naive datetime)
```
django.utils.deprecation.RemovedInDjango20Warning: The MySQL database adapter received an aware datetime (%s), probably from cursor.execute(). Update your code to pass a naive datetime in the database connection's time zone (UTC by default).
```

**Django version 1.8** (requires me to pass an aware datetime)
```
RuntimeWarning: MySQL received a naive datetime (2018-08-08 09:46:22.687201) while time zone support is active.
```
Also here is the log from one of the [travis job](https://travis-ci.org/open-craft/openedx-completion-aggregator/jobs/413493947) when this code was not present.
 

My reason for linking to that blog post, is I could not find a proper link which give details about this. Hence I posted that link because it was giving context that timezone unaware values are being returned.

symbolist: Hmm. Given:

1. https://docs.djangoproject.com/en/2.0/topics/i18n/timezones/#database
2. https://github.com/django/django/commit/e707e4c709c2e3f2dad69643eb838f87491891f8#diff-af003fcfed7cfbdeb396f8647ed0f92fR258
3. https://docs.djangoproject.com/en/2.1/ref/settings/#time-zone

I think we need to do `data.astimezone(timezone.utc).replace(tzinfo=None)` for Django 2+.

@jcdyer What do you think?

symbolist:
@taranjeet Okay, the code is close to done now. You also need to write some tests to check all the new code. In the tests you can use [assertNumQueries](https://docs.djangoproject.com/en/1.11/topics/testing/tools/#django.test.TransactionTestCase.assertNumQueries) to verify the expected number of queries.

completion_aggregator/utils.py
jcdyer: Thanks for explaining @taranjeet.  That's helpful.  

@symbolist: The data *should* already be coming in with tzinfo=utc, but this would be safer, and should be a trivial amount of extra processing time. :+1:

completion_aggregator/tasks/aggregation_tasks.py
```diff
@@ -201,18 +207,35 @@ def update_for_aggregator(self, block, affected_aggregators, force):
...
+            })
+            aggregator = Aggregator(
                 user=self.user,
                 course_key=self.course_key,
                 block_key=block,
                 aggregation_name=block.block_type,
                 earned=total_earned,
                 possible=total_possible,
-                last_modified=last_modified,
+                last_modified=last_modified
```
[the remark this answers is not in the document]
jcdyer: Yep.  The edx python style guide specifies using final commas (See the [third bullet point here](https://edx.readthedocs.io/projects/edx-developer-guide/en/latest/style_guides/python-guidelines.html#syntax-and-organization).  Fewer frustrations when rearranging or adding arguments.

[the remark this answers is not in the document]
taranjeet: Also, it prevents the line being shown in `git diff` next time.

taranjeet:
@symbolist : I have addressed all the review comments and added test using `assertNumQueries` as well. Please have a review :)

completion_aggregator/tasks/aggregation_tasks.py
```diff
@@ -210,7 +216,8 @@ def update_for_aggregator(self, block, affected_aggregators, force):
                 possible=total_possible,
                 last_modified=last_modified,
             )
-            self.aggregators[obj.block_key] = obj
+            self.updated_aggregators.append(aggregator)
+            self.aggregators[block] = aggregator
```
symbolist: I just realized that for blocks with an existing Aggregator object in `self.aggregators`, we should not be creating new objects but instead updating values of the existing objects (without saving them). Otherwise the `created` field will be reset.

completion_aggregator/models.py
```diff
@@ -205,6 +225,21 @@ def __str__(self):
...
 
+    def get_values(self):
+        """
+        Return a dict object containing fields and their values to be used in bulk create or update query.
+        """
+        # these keys values will be assigned as it is
+        exact_keys = ['user', 'course_key', 'block_key', 'aggregation_name', 'earned', 'possible', 'last_modified']
+        values = {}
+        values = {key: getattr(self, key) for key in exact_keys}
+        values['user'] = values['user'].id
```
symbolist: Why not `self.user.id` instead of putting it in `values` first?

completion_aggregator/models.py
```diff
@@ -205,6 +225,21 @@ def __str__(self):
...
+        values['last_modified'] = make_datetime_timezone_unaware(values['last_modified'])
+        values['percent'] = get_percent(values['earned'], values['possible'])
+        values['created'] = make_datetime_timezone_unaware(timezone.now())
+        values['modified'] = make_datetime_timezone_unaware(timezone.now())
```
symbolist: We can't reset the value of created when updating existing `Aggregator` objects. In this method, we should only be reading the values for `self`. You can do instead:

```
values.update({key: make_datetime_timezone_unaware(getattr(self, key)) for key in [
   'last_modified', 'created', 'modified',
]})
```

completion_aggregator/tasks/aggregation_tasks.py
symbolist: Since `pre_save` is not going to get called, `modified` also needs to be updated.

completion_aggregator/models.py
```diff
@@ -205,6 +225,21 @@ def __str__(self):
             percent=self.percent,
         )
 
+    def get_values(self):
+        """
+        Return a dict object containing fields and their values to be used in bulk create or update query.
+        """
+        # these keys values will be assigned as it is
+        exact_keys = ['user', 'course_key', 'block_key', 'aggregation_name', 'earned', 'possible', 'last_modified']
+        values = {}
```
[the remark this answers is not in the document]
taranjeet: @symbolist Sorry, but I didn't understand this comment.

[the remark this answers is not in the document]
symbolist: I meant this is not needed since it is being set in the next line.

completion_aggregator/tasks/aggregation_tasks.py
taranjeet: @symbolist : Please correct me if I am wrong.

I think this is why we are using "INSERT ON DUPLICATE UPDATE" query in mysql. If there will be any object(whose value is already present in the database) and it is being inserted, it will be updated with new values for `earned`, `total`, `percent` and `last_modified`. Another point to support this, when we use `submit_completion`, which internally uses `create_or_update`. It returns a flag which tells if the new object was created or not. In the present codebase, there is no support for such case.

Also since the `Aggregator` objects are not saved, `created` and `modified` will not be populated before being saved to db.

taranjeet:
@symbolist : One more thing, I am not sure that we should write tests for `get_percent` and `get_values`, as they are invoked many times from code which is already being tested. So if anything breaks from these functions, all the tests will start failing.

Also, I was trying to see test cases for other model functions (like block_is_registered_aggregator) or util functions (like get_active_users), but I couldn't find any. Is there any specific reason why test cases are not present for these functions?

completion_aggregator/tasks/aggregation_tasks.py
symbolist: @taranjeet The thing is that `created` and `modified` fields have a default value set to the current time: https://github.com/jazzband/django-model-utils/blob/master/model_utils/fields.py#L12-L36.

Now, if I create a new Aggregator object for an existing `(user, course_key, block_key, aggregation_name)` object, its `created` and `modified` values are not going to be what they are in the database. So the Aggregator objects that are going to be passed to `bulk_create_or_update` and are going to be set in `self.aggregators` are going to be invalid. The difference with `create_or_update` is that the later is passed data with differentiation between the key fields and others and as you noted returns an object that can put into `self.aggregators`. Yes, this does work here because of how `bulk_create_or_update` is internally implemented but callers should not assume anything about that and because they are not currently looking at the `created` or `modified` fields.

As far as I see the only thing needed here is adding a check that if there is an existing object, four of its fields are updated. Otherwise a new object is created as is happening here already.

symbolist:
@taranjeet I do not have historical knowledge of this repo but these should be straightforward to add.

completion_aggregator/tasks/aggregation_tasks.py
taranjeet: @symbolist Thanks for the detailed info. :)

I have updated the codebase.

completion_aggregator/tasks/aggregation_tasks.py
```diff
@@ -201,16 +206,27 @@ def update_for_aggregator(self, block, affected_aggregators, force):
...
-                aggregation_name=block.block_type,
-                earned=total_earned,
-                possible=total_possible,
-                last_modified=last_modified,
-            )
-            self.aggregators[obj.block_key] = obj
+
+            Aggregator.objects.validate(self.user, self.course_key, block)
+            aggregator_identifier_key = '$'.join(map(str, [self.user.id, self.course_key, block, block.block_type]))
+            if aggregator_identifier_key not in self.aggregator_map:
```
symbolist: I don't see why this needs to be so complicated. You can just do `if block not in self.aggregators:`

completion_aggregator/tasks/aggregation_tasks.py
```diff
@@ -201,16 +206,27 @@ def update_for_aggregator(self, block, affected_aggregators, force):
...
+                    last_modified=last_modified,
+                )
+            else:
+                aggregator = self.aggregator_map[aggregator_identifier_key]
+                aggregator.earned = total_earned
+                aggregator.possible = total_possible
+                aggregator.last_modified = last_modified
+            self.aggregator_map[aggregator_identifier_key] = aggregator
+            self.updated_aggregators.append(aggregator)
+            self.aggregators[block] = aggregator
```
symbolist: This can move to the `if` since in the `else` it is already set.

completion_aggregator/tasks/aggregation_tasks.py
taranjeet: I thought I should use what makes an `Aggregator` object unique. Updating this to use `block` only.

completion_aggregator/tasks/aggregation_tasks.py
taranjeet: Got your point. Thanks.

taranjeet:
@symbolist : I have updated the pr as per latest set of review comments. It includes test cases for `get_percent`, `get_values` and a conditional clause to check if aggregator object is present, so its updated rather than being created.

The conditional clause one was a nice catch, which I was thinking that it will be handled totally by "INSERT ON DUP UPDATE" but your comment made more sense about the abstraction part. Thanks :)

completion_aggregator/tasks/aggregation_tasks.py
```diff
@@ -201,16 +205,25 @@ def update_for_aggregator(self, block, affected_aggregators, force):
...
+                    earned=total_earned,
+                    possible=total_possible,
+                    last_modified=last_modified,
+                )
+                self.aggregators[block] = aggregator
+            else:
+                aggregator = self.aggregators[block]
+                aggregator.earned = total_earned
+                aggregator.possible = total_possible
+                aggregator.last_modified = last_modified
```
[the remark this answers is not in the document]
taranjeet: This is addressed. Thanks for pointing it out @symbolist :)

taranjeet:
@symbolist : I have addressed all the review comments. Please review the PR.

symbolist:
- [x] I tested this by turning on completion in LMS and verifying that the Aggregator objects are getting updated correctly when working through a course.
- [x] I read through the code
- [ ] ~~I checked for accessibility issues~~
- [ ] ~~Includes documentation~~

taranjeet:
@symbolist : I have addressed all the review comments and squashed all the commits into a single commit. Hence I am merging this PR.
Thanks for the review and the approval. :)

PR #24, opened 2018-09-14
[OC-5392] Add support for mean value in serializer

xitij2000:
**Description:** 
Fixes bug where mean value was expected by serializer but not present. Added basic support for mean values

**JIRA:** OC-5392

**Testing instructions:**

Check completions via /api/completion/v0/course/gnk/ckn/2018/?requested_fields=mean
There should not be a 500 error. 

**Reviewers:**
- [ ] TBD

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.

**Author concerns:** 
Support for filtering users for mean calculations coming in future PR.

### DIFF
--- a/completion_aggregator/api/v0/views.py
+++ b/completion_aggregator/api/v0/views.py
@@ -327,8 +327,8 @@ class CompletionDetailView(CompletionViewMixin, APIView):
         requested_fields = self.get_requested_fields()
         enrollment = UserEnrollments(self.user).get_course_enrollment(course_key)
         aggregator_queryset = self.get_queryset().filter(
-          course_key=course_key,
-          user=self.user,
+            course_key=course_key,
+            user=self.user,
         )
 
         # Create the list of aggregate completions to be serialized,
--- a/completion_aggregator/api/v0/views.py
+++ b/completion_aggregator/api/v0/views.py
@@ -326,7 +326,10 @@ class CompletionDetailView(CompletionViewMixin, APIView):
             raise NotFound()
         requested_fields = self.get_requested_fields()
         enrollment = UserEnrollments(self.user).get_course_enrollment(course_key)
-        aggregator_queryset = self.get_queryset().filter(course_key=course_key)
+        aggregator_queryset = self.get_queryset().filter(
+          course_key=course_key,
+          user=self.user,
+        )
 
         # Create the list of aggregate completions to be serialized,
         # recalculating any stale completions for this single user.
--- a/completion_aggregator/serializers.py
+++ b/completion_aggregator/serializers.py
@@ -16,6 +16,7 @@ from xblock.core import XBlock
 from xblock.plugin import PluginMissingError
 
 from django.db import transaction
+from django.db.models import Sum
 
 from . import compat
 from .models import Aggregator, StaleCompletion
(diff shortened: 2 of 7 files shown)


completion_aggregator/api/v0/views.py
```diff
@@ -326,7 +326,10 @@ def get(self, request, course_key):
             raise NotFound()
         requested_fields = self.get_requested_fields()
         enrollment = UserEnrollments(self.user).get_course_enrollment(course_key)
-        aggregator_queryset = self.get_queryset().filter(course_key=course_key)
+        aggregator_queryset = self.get_queryset().filter(
```
xitij2000: I'm not sure why we need to fetch the enrollment here and use the ``course_id`` and ``user`` from that, would those not be the same as ``course_key`` and ``self.user``?

jcdyer: Only if aggregators exist.  If the user hasn't interacted with the course, we still need to report that they have made 0% progress.  I believe @pomegranited came up with this strategy.

jcdyer: Oh.  I just noticed that this is on the detail view.  I was thinking of different code on the list view, to get results for all users in the course.

And that we've already checked that the enrollment exists right above.  So I'm not sure why it's here.

completion_aggregator/serializers.py
```diff
@@ -255,6 +256,19 @@ def get_username(self, obj):
...
+        """
+        Return the mean completion percent for all enrolled users.
+        """
+        enrollments = compat.get_users_enrolled_in(obj.course_key)
+        enrollment_count = enrollments.count()
+
+        mean = Aggregator.objects.filter(
+            course_key=obj.course_key,
+            aggregation_name='course',
+        ).aggregate(Sum('percent')).get('percent__sum', 0) / enrollment_count
```
bradenmacdonald: We should probably check first to make sure `enrollment_count` is not zero, even though that should be rare.

xitij2000: Done, I'm returning 0 in that case.

tests/test_serializers.py
```diff
@@ -253,6 +253,36 @@ def test_zero_possible(self):
...
+                user=self.test_user,
+                course_key=course_key,
+                block_key=course_key.make_usage_key(block_type='course', block_id='course{}'.format(idx)),
+                aggregation_name='course',
+                earned=data[idx][0],
+                possible=data[idx][1],
+                last_modified=timezone.now(),
+            )[0]
+            for idx in range(4)
+        ]
```
jcdyer: Nit: `for datum in data` works here.

xitij2000:
@jcdyer The original API was caching this using regular Django caching. Should I add that here? There will be another PR to change this and allow role-based filtering of means soon (next sprint), so there probably should be some caching.

tests/test_serializers.py
xitij2000: @jcdyer  I need a number to give each block a different id, hence this approach idx.

jcdyer:
Works for me. Even a short cache would resolve the problem, and since Aggregators are only updated hourly anyway, it shouldn't even be noticeable.

tests/test_serializers.py
jcdyer: Oh.  I looked for something like that, but missed it.

PR #27, opened 2018-09-26
Fix error when getting means for course without aggregators

xitij2000:
**Description:** 

Fixes error that occurs when you ask for mean value in a course without any Aggregators. 

**Reviewers:**
- [ ] @jcdyer 

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.

### DIFF
--- a/completion_aggregator/serializers.py
+++ b/completion_aggregator/serializers.py
@@ -16,7 +16,8 @@ from xblock.core import XBlock
 from xblock.plugin import PluginMissingError
 
 from django.core.cache import cache
-from django.db.models import Sum
+from django.db.models import Sum, Value
+from django.db.models.functions import Coalesce
 
 from . import compat
 from .models import Aggregator, StaleCompletion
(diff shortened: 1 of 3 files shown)


completion_aggregator/serializers.py
```diff
@@ -259,22 +259,32 @@ def get_username(self, obj):
...
+        """
+        enrollments = compat.get_users_enrolled_in(obj.course_key)
+        enrollment_count = enrollments.count()
+        if enrollment_count == 0:
+            return 0.
+
+        total = Aggregator.objects.filter(
+            course_key=obj.course_key,
+            aggregation_name='course',
+        ).aggregate(Sum('percent')).get('percent__sum', 0.)
```
jcdyer: I'm not sure this is right.  I think if there are any nulls, the entire sum will become 0.  I suspect we need a more complex sql expression for the aggregate? Something like `Sum(Coalesce('percent', Value(0)))`.

[Database function documentation](https://docs.djangoproject.com/en/2.1/ref/models/database-functions/)

xitij2000: OK! Didn't know that. Will update the PR and add a test for this.

xitij2000: I've updated the code, but I don't think it's possible for percent to be null, so I'm not sure it's possible to test this.

completion_aggregator/serializers.py
```diff
@@ -271,10 +272,10 @@ def _calculate_mean(self, obj):
         total = Aggregator.objects.filter(
             course_key=obj.course_key,
             aggregation_name='course',
-        ).aggregate(Sum('percent')).get('percent__sum', 0.)
+        ).aggregate(
+            total=Sum(Coalesce('percent', Value(0.)))
+        ).get('total') or 0.
```
jcdyer: Looks cleaner too.

PR #34, opened 2018-10-12
Add options to mgt command.

jcdyer:
**Description:** Describe in a couple of sentence what this PR adds

**JIRA:** Link to JIRA ticket

instructions.

**Testing instructions:**

1. Open page A
2. Do thing B
3. Expect C to happen
4. If D happend instead - check failed.

**Reviewers:**
- [ ] tag reviewer 

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.

migrations, etc.

### DIFF
--- a/completion_aggregator/aggregator.py
+++ b/completion_aggregator/aggregator.py
@@ -10,10 +10,9 @@ from __future__ import absolute_import, division, print_function, unicode_litera
 
 import collections
 import logging
+import time
 
 import six
-
-from django.db import transaction
 
 from . import models, utils
 from .tasks import aggregation_tasks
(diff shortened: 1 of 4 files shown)


jcdyer:
Fyi @edX/devops @feanil. Added options to limit the number of stale completions processed in an hour (suggest a number like 100,000 to 1,000,000 for that, if that works with memory constraints it needs to be high enough to easily handle the number of completions that come in during an hour), also added an Option to limit the number of rows pulled from the db at a time (10,000 might be good) and a routing key argument as well. Let me know if you need anything else.

jcdyer:
The failing test is a trivial change in argument syntax. Power is currently out at my house, so I'll fix that in the morning.

completion_aggregator/aggregator.py
```diff
@@ -22,22 +23,41 @@
...
+    limit = min(limit, initial_count)
+    enqueued = 0
+    task_options = {}
+    if routing_key:
+        task_options['routing_key'] = routing_key
+    while enqueued < limit:
         stale_blocks = collections.defaultdict(set)
         forced_updates = set()
-        for stale in stale_queryset:
+        evaluated = stale_queryset[:batch_size]  # This actually works fine if batch_size is None
```
bradenmacdonald: If `batch_size` is say 100, isn't this going to keep processing the same 100 entries over and over? I don't see anything that forces the query to restart.

Also, if some other process handled a few StaleCompletions while this ran, is it possible this could become an infinite loop, because `len(evaluated)` could be 0 and `enqueued` will never reach `limit`?

completion_aggregator/management/commands/run_aggregator_service.py
```diff
@@ -24,12 +24,43 @@ class Command(BaseCommand):
...
+        )
+        parser.add_argument(
+            '--delay-between-batches',
+            help='Amount of time to wait between processing batches in seconds.  (default: 0.0)',
+            default=0.0,
+            type=float,
+        )
+        parser.add_argument(
+            '--limit',
+            help='Amount of time to wait after submitting each 1000 tasks.  (default: 0.0)',
```
bradenmacdonald: This help text is wrong and the `limit` parameter doesn't seem to be used - it doesn't get passed to `perform_aggregation`

completion_aggregator/aggregator.py
```diff
@@ -22,22 +23,41 @@
...
+    if limit is None:
+        limit = float('inf')
+    limit = min(limit, initial_count)
+    enqueued = 0
+    task_options = {}
+    if routing_key:
+        task_options['routing_key'] = routing_key
+    while enqueued < limit:
         stale_blocks = collections.defaultdict(set)
         forced_updates = set()
```
bradenmacdonald: If these vars are going to be used outside of this loop, they should be declared outside of the loop

completion_aggregator/aggregator.py
```diff
@@ -22,22 +23,41 @@
...
     Enqueues tasks to reaggregate modified completions.
 
     When blocks are completed, they mark themselves as stale.  This function
     collects all stale blocks for each enrollment, and enqueues a single
     recalculation of all aggregators containing those stale blocks.
+
+    batch_size (int|None):
+        Maximum number of stale completions to enqueue in a single run.
+        None = Process all outstanding StaleCompletions
+
```
bradenmacdonald: Missing doc for the `limit` param

completion_aggregator/aggregator.py
jcdyer: Yup.  Updated to use a pk-based indexing scheme.

completion_aggregator/aggregator.py
jcdyer: Yes they should.

completion_aggregator/aggregator.py
jcdyer: Adding.

bradenmacdonald:
Testing notes:

First I generated a bunch of `StaleCompletion` entries:

```python
from django.core.management.base import BaseCommand
from django.db import transaction

from opaque_keys.edx.keys import CourseKey, UsageKey
from completion_aggregator.models import StaleCompletion

class Command(BaseCommand):
    def handle(self, *args, **options):
        for u in range(0, 100):
            for c in range(0, 100):
                with transaction.atomic():
                    for b in range(0, 100):
                        username = 'user{:02d}'.format(u)
                        course_key = CourseKey.from_string('course-v1:OpenCraft+testing+R{:02d}'.format(c))
                        block_key = UsageKey.from_string('block-v1:OpenCraft+testing+only+type@vertical+block@00112233445566778899aabbccddee{:02d}'.format(b))
                        StaleCompletion.objects.create(username=username, course_key=course_key, block_key=block_key)
            print('Created {} of 100 users'.format(u + 1))
```

Then with the old branch, I put a breakpoint just before `aggregation_tasks.update_aggregators.delay` and then ran `./manage.py lms run_aggregator_service --settings=devstack_docker`. The process's memory usage climbed above 1GB and I killed it at that point to spare my poor devstack. It never reached the breakpoint.

Then I checked out this new branch and ran the exact same command. This time, the memory usage stayed <400MB and the breakpoint was successfully reached.

jcdyer:
Squashed commits.  Waiting on tests.

PR #36, opened 2018-10-17
Add indexes to StaleCompletion table

jcdyer:
**Description:** This adds indexes to the stale completion table.  They cannot be added the simple way, because that locks the table for multiple minutes.  So instead, we perform the following steps:

1. Create a new table with the indexes.
2. Copy the data to the table in batches (up to a known maximum PK).
3. Rename the tables. 
4. Copy any data with a PK higher than the original maximum, again in batches, just to be safe.

Batch size is set to 1000 rows, which makes each batch take about half a second, which keeps lock contention low.

This has been tested on a data set of 1,000,000 rows, and took on the order of 8 minutes.  , so it will probably take around four hours in staging, but will not hold a lock on the tables during that time.

Note: The procedure for migrating the data without downtime was cribbed freely from the excellent Percona tools: https://www.percona.com/doc/percona-toolkit/2.2/pt-online-schema-change.html

**JIRA:** YONK-1175

**Dependencies:** 

**Merge deadline:** ASAP

**Testing instructions:**

Verify that the index does its job.  querying the completion_aggregator_stalecompletion table WHERE username=... AND course_key=... AND resolved=False AND created < NOW() does not perform a full table scan. 

Verify that the migration runs, adds indexes to the table, and does not prevent data from being added to the table while it is running:

```sql
INSERT INTO completion_aggregator_stalecompletion 
    (created, modified, username, course_key, block_key, `force`, resolved)
    VALUES 
        (NOW(), NOW(), "username", "COURSE/COURSE/COURSE", "i4x:COURSE/COURSE/html/BLOCK", FALSE, FALSE)
```

should be sufficient to create a record while the migration is running.

**Reviewers:**
- [x] @UmanShahzad 

**Merge checklist:**
- [x] All reviewers approved
- [x] CI build is green
- [x] Version bumped
- [ ] Changelog record added
- [x] Documentation updated (not only docstrings)
- [x] Commits are squashed
- [x] PR author is listed in AUTHORS

**Post merge:**
- [ ] Create a tag
      finished.


tox.ini
```diff
@@ -28,7 +28,7 @@ deps =
     django20: Django>=2.0,<2.1
     -r{toxinidir}/requirements/test.txt
 commands =
-    py.test {posargs}
+    py.test --nomigrations {posargs}
```
jcdyer: pytest doesn't like running model code during migrations, so it chokes on L34 of the migration.  There's probably a better way to resolve it than skipping migrations, but that is lower priority than this PR.

jcdyer:
Note to reviewer:  to test this, just create stale completion objects directly in the db. Stale Completion.objects.bulk_create() and loop over 100 courses 1000 users, and 5 block keys per course (coursekey.make_usage_key("html", "block%s" % I) will help).

completion_aggregator/migrations/0004_index_stalecompletion.py
```diff
@@ -0,0 +1,83 @@
...
+            SELECT `created`, `modified`, `username`, `course_key`, `block_key`, `force`, `resolved`
+                FROM %(source)s
+                WHERE id > %%s
+                ORDER BY id
+                LIMIT %%s OFFSET %%s
+    """
+    StaleCompletion = apps.get_model("completion_aggregator", "StaleCompletion")
+    cursor = schema_editor.connection.cursor()
+    max_id = StaleCompletion.objects.order_by('-id')[0].id
+    count = StaleCompletion.objects.count()
```
UmanShahzad: If anything gets added before this line can run but after we already have a `max_id`, then when trying to insert any new data that came in while performing the initial copy, we'll end up trying to insert data that already exists in the table, namely data with `id > max_id AND id < count`. Is that going to cause an issue? If so we can run these 2 as one transaction.

completion_aggregator/migrations/0004_index_stalecompletion.py
```diff
@@ -0,0 +1,83 @@
...
+        cursor.execute(copy_sql, [0, BATCH_SIZE, offset])
+
+    cursor.execute(
+        """
+        RENAME TABLE
+            completion_aggregator_stalecompletion TO completion_aggregator_stalecompletionold,
+            completion_aggregator_stalecompletionnew TO completion_aggregator_stalecompletion
+        """
+    )
+    end_count = StaleCompletion.objects.filter(id__gt=max_id).count()
```
UmanShahzad: After renaming tables, wouldn't this look inside the table that we just created, which would technically have no ID greater than `max_id` (which could still happen from what I mentioned in the comment above, but even then it'd be missing some things)?

UmanShahzad:
@jcdyer Thanks, the latest diff looks good. I'll try to test it locally though if you've already tested it on a 1 million row table, I might not be able to do anything more special.

UmanShahzad:
@jcdyer Sounds good.

jcdyer:
Separated the db operations from the state transitions, and tested that they end up in a sane state with `./manage.py lms --settings=devstack makemigrations -e` (the command that failed in the platform tests).

jcdyer:
Please also give a thumb on https://github.com/edx-solutions/edx-platform/pull/1199.

completion_aggregator/migrations/0004_index_stalecompletion.py
```diff
@@ -0,0 +1,102 @@
...
+        'source': 'completion_aggregator_stalecompletion',
+        'target': 'completion_aggregator_stalecompletionnew',
+    }
+    for offset in six.moves.range(0, count, BATCH_SIZE):
+        cursor.execute(copy_sql, [0, BATCH_SIZE, offset])
+
+    cursor.execute(
+        """
+        RENAME TABLE
+            completion_aggregator_stalecompletion TO completion_aggregator_stalecompletionold,
```
UmanShahzad: I noticed we don't drop the old table after this migration is over. Is that intended? Is it something we'll come back to later?

jcdyer: Yes.  I don't want to delete it until we can do some manual post-verification.

UmanShahzad: Sounds good.

PR #42, opened 2018-11-01
BB-403 Course level completion view

jghyllebert:
**Description:**
This PR allows to get the mean completion for a course.

**JIRA:** https://tasks.opencraft.com/browse/BB-403

**Dependencies:** None

**Merge deadline:** None

**Installation instructions:** None
**Testing instructions:**
TBD
**Reviewers:**
- [x] tag reviewer 

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.

**Author concerns:**
I'm not a big fan of `user_has_excluded_roles`, but mocking `CourseAccessRole` from edx-platform is not available.
`user_has_excluded_roles` basically only exists for mocking reasons.

Will edit version after successful review


completion_aggregator/api/v1/views.py
```diff
@@ -388,3 +390,109 @@ def get(self, request, course_key):
...
+                # the results.
+                continue
+            if 'staff' in roles_to_exclude and enrolled_user.is_staff:
+                # Double check in case staff role is handled by Django
+                # User is a staff member and staff needs to be excluded
+                continue
+            cohort_id = cohorts.get_user_cohorts(enrolled_user)
+            if cohort_id == cohort_filter:
+                # Keep user in the results
+                filtered_enrollments.append(enrollment)
```
jcdyer: With tens of thousands of users in a course, this loop is going to be massively slow.  We need to filter the enrollments in the database.  Let me know if you need pointers on making more complex queries using the ORM.

tests/test_views.py
```diff
@@ -557,6 +561,106 @@ def test_detail_view_staff_all_users(self, mock_update):
...
+            possible=8.0,
+            last_modified=timezone.now(),
+        )
+
+        response = self.client.get(self.get_course_stat_url(
+            'edX/toy/2012_Fall',
+            cohorts=1,
+            exclude_roles='beta'
+        ))
+        self.assertEqual(excluded_roles_mock.call_count, 2)
```
jcdyer: Is this getting called twice because there are two users enrolled in the course?  that's going to be a performance bottleneck in large courses.  We want to make sure we're doing a constant number of queries regardless of the course size.

tests/test_views.py
```diff
@@ -557,6 +561,106 @@ def test_detail_view_staff_all_users(self, mock_update):
...
+
+        response = self.client.get(self.get_course_stat_url(
+            'edX/toy/2012_Fall',
+            cohorts=1,
+            exclude_roles='staff'
+        ))
+
+        self.assertEqual(response.data['results'][0]['completion']['possible'], 8.0)
+        self.assertEqual(response.data['results'][0]['completion']['earned'], 3.4)
+        self.assertEqual(response.data['results'][0]['completion']['percent'], 0.425)
```
jcdyer: That's an unexpectedly clean fraction!

completion_aggregator/api/v1/views.py
```diff
@@ -388,3 +390,109 @@ def get(self, request, course_key):
...
+            earned=Avg('earned'),
+            percent=Sum('earned') / Sum('possible'))
+        completions['course_key'] = course_key
+
+        # Return the paginated, serialized completions
+        serializer = self.get_serializer_class()(
+            instance=[completions],
+            requested_fields=requested_fields,
+            many=True
+        )
```
jcdyer: This view should define its own serializer.  You asked about this on the ticket, and I'm sorry I wasn't able to respond sooner.  

The other views are effectively returning a list of aggregators for users.  This view is returning a collection of statistics about the course (currently the collection is only a single statistic -- the mean completion, but imagine we add other stats in the future, so the response looks like:

    {
        "results": [{
            "mean_completion": {
                "earned": 45.0, 
                "possible": 100.0, 
                "percent": 0.45 
            },
            "max_completion": {
                "earned": 98.0, 
                "possible": 100.0, 
                "percent": 0.98 
            },
            "most_complete_cohort": 3
        }]
    }

And don't forget to include the envelope information ("course_id", "filters") in the response body.

completion_aggregator/api/v1/views.py
```diff
@@ -388,3 +390,109 @@ def get(self, request, course_key):
             many=True
         )
         return paginator.get_paginated_response(serializer.data)
+
+
+class CourseLevelCompletionView(CompletionViewMixin, APIView):
```
jcdyer: This name is semantically pretty similar to what the CompletionDetailView returns.  Can you include `Stats` in the name, so the distinction is clearer?

.gitignore
```diff
@@ -76,3 +76,6 @@ tests/__init__.py
 
 # Development task artifacts
 default.db
+
+# Virtual environments
+.env
```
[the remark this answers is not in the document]
jghyllebert: I was using pipenv as I didn't know piptools is the tool we're using. Pipenv has builtin support for .env files.

I probably shouldn't have put Virtual environments as the comment though.

completion_aggregator/api/v1/views.py
jcdyer: Maybe push the query refinements into methods that can be mocked.

Something like:
```python
aggregator_qs = self.get_queryset(course_key, enrollments)
if roles_to_exclude:
    self.exclude_roles(roles_to_exclude)
...

def get_initial_queryset(course_key, enrollments):
    return self.get_queryset().filter(course_key=course_key, user__in enrollment.user for enrollment in paginated_enrollments)

def exclude_roles(self, queryset, roles_to_exclude):
     if roles_to exclude:
        return queryset.exclude(user__courseaccessrole__role__in=roles_to_exclude)
...
```
And then mock get_initial_queryset to return a list, and mock exclude_users to filter that list.

jcdyer: (Response made without looking at the latest updates to the code.  Checking it out now.

completion_aggregator/api/v1/views.py
```diff
@@ -388,3 +390,124 @@ def get(self, request, course_key):
...
+        course_key = CourseKey.from_string(course_key)
+        paginator = self.pagination_class()  # pylint: disable=not-callable
+        requested_fields = self.get_requested_fields()
+        cohorts = UserCohorts(course_key)
+        course_cohorts = cohorts.get_course_cohorts()
+        if not course_cohorts:
+            raise NotFound()
+        roles_to_exclude = self.request.query_params.get('exclude_roles', '').split(',')
+        cohort_filter = self.request.query_params.get('cohorts')
+        if cohort_filter and not isinstance(int, cohort_filter):
```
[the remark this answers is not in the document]
malinoff: I just realized that I started the review but did not submit it at the time after I wrote this comment :(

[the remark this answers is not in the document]
malinoff: @jcdyer actually covered this problem, so I'm going to resolve this one.

completion_aggregator/api/v1/views.py
```diff
@@ -388,3 +393,136 @@ def get(self, request, course_key):
...
+            later.
+
+        exclude_roles (optional):
+            A comma separated list of roles to exclude from the results.
+
+    **Returns**
+
+        * 200 on success with above fields.
+        * 400 if an invalid value was sent for requested_fields and cohorts,
+        * 404 If the course is not cohorted
```
jcdyer: It needs to return a 200 with the results for all users in the course (minus excluded roles) if the course is not cohorted (unless a cohort ID is provided, in which case it will return a 400).

completion_aggregator/api/v1/views.py
```diff
@@ -388,3 +393,136 @@ def get(self, request, course_key):
...
+                        "mean_completion": {
+                            "earned": 3.4,
+                            "possible": 8.0,
+                            "percent": 0.425
+                        }
+                    }
+                ]
+            }
+
+    """
```
jcdyer: You mentioned you were having trouble making the response turn out in this exact format.  I haven't tested manually yet, but please make sure the docs are up to date.

completion_aggregator/serializers.py
```diff
@@ -282,6 +282,65 @@ def get_mean(self, obj):
...
+        super(CourseCompletionStatsSerializer, self).__init__(
+            instance, *args, **kwargs)
+        for field in self.optional_fields - requested_fields:
+            del self.fields[field]
+
+    def get_username(self, obj):
+        """
+        Serialize the username.
+        """
+        return obj.user.username
```
jcdyer: I don't think there's any need for a username on this serializer.  In fact, the username won't be available, because it's just fetching aggregate results.

completion_aggregator/api/v1/views.py
```diff
@@ -388,3 +393,136 @@ def get(self, request, course_key):
...
+        roles_to_exclude = self.request.query_params.get('exclude_roles', '').split(',')
+        cohort_filter = self._parse_cohort_filter(
+            self.request.query_params.get('cohorts'))
+
+        enrollments = UserEnrollments().get_course_enrollments(course_key)
+        paginated = paginator.paginate_queryset(
+            enrollments, self.request, view=self)
+        aggregator_qs = self.get_queryset().filter(
+            course_key=course_key,
+            user__in=[enrollment.user for enrollment in paginated])
```
jcdyer: The queryset will also need to filter by `aggregation_name='course'`, otherwise the sums and ratios will get all wonky.  Each course has an overall completion aggregator (course), plus a subdivided aggregator for each chapter, and for each section within a chapter (sequential), and for each unit (vertical) below that.

completion_aggregator/api/v1/views.py
```diff
@@ -388,3 +393,136 @@ def get(self, request, course_key):
...
+        if roles_to_exclude:
+            aggregator_qs = aggregator_qs.exclude(
+                user__courseaccessrole__role__in=roles_to_exclude)
+        if cohort_filter is not None:
+            aggregator_qs = aggregator_qs.exclude(
+                user__cohortmembership__course_user_group__pk=cohort_filter)
+        completions = aggregator_qs.aggregate(
+            possible=Avg('possible'),
+            earned=Avg('earned'),
+            percent=Sum('earned') / Sum('possible'))
```
jcdyer: Nit, rename this `completion_stats`.  `completions` sounds like a collection of Aggregator or BlockCompletion objects, but since we've aggregated (in the database sense) we no longer have a collection.

completion_aggregator/serializers.py
```diff
@@ -282,6 +282,65 @@ def get_mean(self, obj):
...
+            return 0.
+
+        total = Aggregator.objects.filter(
+            course_key=obj.course_key,
+            aggregation_name='course',
+        ).aggregate(
+            total=Sum(Coalesce('percent', Value(0.)))
+        ).get('total') or 0.
+
+        return total / enrollment_count
```
jcdyer: All this work is already performed in the view.  It looks like you're throwing it out and redoing the calculation, but without the filtering.  At this point, `mean` should just be available as a field on `obj`.

tests/test_views.py
```diff
@@ -557,6 +561,119 @@ def test_detail_view_staff_all_users(self, mock_update):
...
+    @XBlock.register_temp_plugin(StubSequential, 'sequential')
+    @XBlock.register_temp_plugin(StubHTML, 'html')
+    def test_stat_view_course_no_cohorts(self):
+        response = self.client.get(self.get_course_stat_url(
+            'edX/toy/2012_Fall',
+            cohorts=1,
+            exclude_roles='staff',
+        ))
+
+        self.assertEqual(response.status_code, 200)
```
jcdyer: Ah.  I see it's doing the right thing with no cohort provided.  Then the above doc fix should just be in the docs. :)

jcdyer:
@malinoff This mostly looks good, but I'm afraid there are two issues remaining.  Hopefully they'll be easy to fix, but they are definitely showstoppers for being able to use this in production as-is.

I discovered them both by having a test course with 4000+ users enrolled in it.

**First:** The query is making a separate call to get each user after getting the enrollments.  I actually made this same mistake recently.  The fix is either to replace 

```diff
    - user__in=[enrollment.user for enrollment in enrollments])                                             
    + user_id__in=[enrollment.user_id for enrollment in enrollments])                                             
```

on views.py, L519, or to add a `.select_related('user')` to the enrollments queryset.

**Second:** The query is returning the average completion of users *who have engaged with the course*.  Before a user interacts with the course , they don't have any aggregators.  We will need to treat these users as if they have a completion of 0.0 / ???.   It is technically possible for the denominator to be different for different users, but I don't think we need to worry about that.  Using the average `possible` value for the users that exist, and multiplying it by the number of enrollments is good enough.  So something like:

```python
    earned=Sum('earned') / len(enrollments),
    possible=Avg('possible'),
    percent=Sum('earned') / (Avg('possible') * len(enrollments))
```

should work, if you also filtering out cohorts and roles on the `enrollments` queryset instead of on `aggregators`.

jcdyer:
It's got my approval now.  @malinoff Please look at the latest commit, and make sure you're happy with it, then squash and merge.

jcdyer:
@malinoff FYI: I also fixed a mismatch between the name of the mean_completion variable in the response docs vs. the actual response.

PR #43, opened 2018-11-03
BB-524: Handle stale completions for unknown users.

jcdyer:
**Description:** 

When updating aggregators, sometimes a StaleCompletion exists for a User that doesn't exist (probably from old data or test data that sneaked in).  Those cause harmless errors, but the errors mean that the StaleCompletion doesn't get resolved, so the same errors occur every hour.  This PR resolves the StaleCompletions.

**JIRA:** [BB-524](https://tasks.opencraft.com/browse/BB-524)

**Merge deadline:** None

instructions.

**Testing instructions:**

1. Create a stale completion for a non-existent user.
2. Run `./manage.py run_aggregator_service`
3. Verify that the StaleCompletion is marked resolved, and that no exceptions are raised.

Note, this goes through a celery task so make sure CELERY_ALWAYS_EAGER is enabled, or that workers are running.

**Reviewers:**
- [ ] TBD

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.

**Author concerns:** I also included some general bookkeeping and cleanup into this PR, but kept them in a separate commit.  It may be easier to review this PR one commit at a time.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,11 +40,11 @@ docs: ## generate Sphinx HTML documentation, including API docs
 
 upgrade: ## update the requirements/*.txt files with the latest packages satisfying requirements/*.in
 	pip install -q pip-tools
-	pip-compile --upgrade -o requirements/dev.txt requirements/base.in requirements/dev.in requirements/quality.in
-	pip-compile --upgrade -o requirements/doc.txt requirements/base.in requirements/doc.in
-	pip-compile --upgrade -o requirements/quality.txt requirements/quality.in
-	pip-compile --upgrade -o requirements/test.txt requirements/base.in requirements/test.in
-	pip-compile --upgrade -o requirements/travis.txt requirements/travis.in
+	pip-compile --upgrade --no-index -o requirements/dev.txt requirements/base.in requirements/dev.in requirements/quality.in
+	pip-compile --upgrade --no-index -o requirements/doc.txt requirements/base.in requirements/doc.in
+	pip-compile --upgrade --no-index -o requirements/quality.txt requirements/quality.in
+	pip-compile --upgrade --no-index -o requirements/test.txt requirements/base.in requirements/test.in
+	pip-compile --upgrade --no-index -o requirements/travis.txt requirements/travis.in
 	# Let tox control the Django version for tests
 	sed '/^django==/d' requirements/test.txt > requirements/test.tmp
 	mv requirements/test.tmp requirements/test.txt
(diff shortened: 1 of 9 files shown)


requirements/dev.txt
```diff
@@ -4,6 +4,7 @@
 #
 #    pip-compile --output-file requirements/dev.txt requirements/base.in requirements/dev.in requirements/quality.in
 #
+--index-url http://localhost:3141/root/pypi/+simple/
```
UmanShahzad: @jcdyer I think you have to add --no-index to avoid this.

jcdyer: Thanks.

requirements/dev.txt
```diff
@@ -2,82 +2,93 @@
...
-packaging==17.1           # via caniusepython3, tox
-path.py==11.0.1           # via edx-i18n-tools
-pbr==4.1.0                # via stevedore
-pip-tools==2.0.2
+path.py==11.5.0           # via edx-i18n-tools
+pbr==5.1.0                # via stevedore
+pip-tools==3.1.0
 pkginfo==1.4.2            # via twine
-pluggy==0.6.0             # via tox
+pluggy==0.8.0             # via tox
```
xitij2000: Make requirements failed for me because this file wants pluggy 0.8 while test.txt wants 0.6. 
Switching to pytest==3.7.x should resolve this I think.

jcdyer: So switching to a newer pytest helps this issue, but creates a new issue in that the test suite starts complaining about django.conf.urlresolvers being deprecated, but we need to use it to continue to support django 1.8.  I'll see if I can silence those warnings with a conditional import, and then upgrade pytest again.

jcdyer: Resolving that lead to a complaint about pytest fixtures.  And then on to a complaint about not being able to find contextlib (only needed for python 2).  I tried building all the requirements from python 2 instead of 3, but that lead to another set of errors.

jcdyer:
I removed the requirement updates, because they were starting to prove complicated.  It's getting hard to support the same set of requirements on python 2 and 3 using pip compile.  I'll tackle that in a separate someday PR.

jcdyer:
No version bump on this one.  It's low priority, so I think it's fine to let it sit until there's a higher priority ticket worth a version bump.

PR #44, opened 2018-11-06
Add root_block option to only get a subtree of aggregators

jcdyer:
**Description:** Adds the ability to specify a block key to serve as the root of the requested aggregators.  If the view would calculate them on the fly, it instead only calculates the aggregators under (and including) the specified block.  

Some notes: 
* if it doesn't calculate the aggregators, it ignores the root_block parameter, and sends back all the data, because it would take more time to calculate which ones need to be returned than just to make the single indexed db query to get the blocks.  
* One point of ugliness is that the course-level aggregator gets returned as `0.0/None/0.0`  I don't see an easy way to resolve that, given the format of the response, but since the user only wanted a subtree anyway, we'll expect them not to use that data.

**JIRA:** https://tasks.opencraft.com/browse/BB-538

**Dependencies:** N/A

**Merge deadline:** ASAP

**Testing instructions:**

1. Visit an the aggregator detail page for a full-size course, specifying `?username=<yourusername>&requested_fields=chapter,sequential,vertical`
2. Verify that it returns all blocks for the course.  Check the response time in the browsers dev tools.
3. Add `&root_block=<a chapter block key>`.  
4. Verify that it returns fewer blocks, the course completion is empty `0.0/None/0.0`, and that performance improved.

**Reviewers:**
- [ ] tag reviewer 

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.

migrations, etc.

### DIFF
--- a/completion_aggregator/api/v1/views.py
+++ b/completion_aggregator/api/v1/views.py
@@ -6,11 +6,12 @@ from __future__ import absolute_import, division, print_function, unicode_litera
 
 from collections import defaultdict
 
-from opaque_keys.edx.keys import CourseKey
+from opaque_keys.edx.keys import CourseKey, UsageKey
 from rest_framework.exceptions import NotFound
 from rest_framework.views import APIView
 
 from ... import compat
+from ...models import StaleCompletion
 from ...serializers import AggregatorAdapter
 from ..common import CompletionViewMixin, UserEnrollments
 
@@ -264,6 +265,9 @@ class CompletionDetailView(CompletionViewMixin, APIView):
             all enrolled users is returned. If the requesting user does not have
             staff access, it will return a 403 Error.
 
+        root_block (optional):
+            Get aggregators under a certain block, not for the whole course.
+
         requested_fields (optional):
             A comma separated list of extra data to be returned.  This can be
             one of the block types specified in `AGGREGATE_CATEGORIES`.  If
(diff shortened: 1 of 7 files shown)


completion_aggregator/api/v1/views.py
```diff
@@ -351,10 +355,16 @@ def get(self, request, course_key):
...
             recalculate_stale = False
+            is_stale = False
         else:
             if not UserEnrollments(self.user).is_enrolled(course_key):
                 # Return 404 if effective user does not have an active enrollment in the requested course
                 raise NotFound()
+            is_stale = StaleCompletion.objects.filter(
+                username=self.user.username,
+                course_key=course_key,
+                resolved=False,
```
UmanShahzad: Noting that the multi-field index on this table has `created` before `resolved`, so we'll end up doing a full scan of stale completions of a user in a course. I have no idea how big those set of rows look like, but I imagine not big when filtered down that much already, so probably nothing to do extra here.

jcdyer: Yeah, the resolved StaleCompletions get deleted every hour, so there's not much extra work that needs to be done.  I'll take another look at indexes when things calm down, but it won't be a bottleneck.

jcdyer:
Specific URLs to test for this PR:

http://lms.mcka.local/api/completion-aggregator/v1/course/edX/PS1/20181029/?username=cliff2&root_block=i4x%3A%2F%2FedX%2FPS1%2Fchapter%2F4bc4a183eddac3b4792a645a8742b618&requested_fields=chapter%2Csequential%2Cvertical&page_size=200

Replace with your own username, course_key and root_block for a chapter in your course.

Test with and without the root_block specifier.

Make sure you test with a StaleCompletion present for the course.  `completion_aggregator.models.StaleCompletion.objects.filter(username='mslm', course_key=opaque_keys.edx.keys.CourseKey.from_string('your-course-v1:a+b+c'), resolved=False).exists()`

If it doesn't exist, you can create one like: 

    StaleCompletion.objects.create(
        username='mslm', 
        course_key=CourseKey.from_string(...), 
        force=False, 
        resolved=False)

PR #46, opened 2018-11-15
[MCKIN-8780] Prefetch the auth_user table data since we eventually use it.

UmanShahzad:
Quick optimization fix - we're currently making 1 query per user because we eventually call into enrollment.user without having that data available. This will make that data available without extra queries.


completion_aggregator/api/v1/views.py
```diff
@@ -371,7 +371,7 @@ def get(self, request, course_key):
 
         if 'user_ids' in request.query_params:
             user_ids = (int(id) for id in request.query_params['user_ids'].split(','))
-            enrollments = enrollments.filter(user_id__in=user_ids)
+            enrollments = enrollments.filter(user_id__in=user_ids).select_related('user')
```
jcdyer: This select_related looks like it should be applied even if the `user_ids` parameter is not provided.  Please move this up to L355 and L370 (or apply it once before L372.

UmanShahzad: @jcdyer Would L376 work?

UmanShahzad: Going with that.

jcdyer:
Approved!

PR #49, opened 2018-12-19
MCKIN-8827: Cache updater blockstructure data.

jcdyer:
**Description:** 

I discovered by profiling (using snakeviz) that the majority of time is spent generating the coursestructure, and in particular, processing the `StudentViewTransformer`.  So I cache it.  

The one complication is that sometimes we need to invalidate the cache for all the users in a course (and users can have a cached block structure for a partial course as well), so we needed to add a mechanism to treat all those caches as invalidated as well.  Rather than explicitly deleting those cache entries, I just skip retrieving them, so the user will grab a fresh copy and cache that.  

NOTE: @bradenmacdonald said there should already be some caching going on for the block structure.  If anyone knows why that's getting bypassed here, I'm totally open to tearing this code out.

**JIRA:** MCKIN-8827, BB-743

**Dependencies:** N/A

**Merge deadline:** Jan. 3.

**Testing instructions:**  Visit a few modules in Apros.  Verify that after the first one, page loads (and in particular, calls to the aggregator) get snappier after the first one, and that entries get cached.

You can replicate my profiling by doing the following:

Browse an uncompleted section of a course to create new StaleCompletions.  Then run the following with the master branch of openedx-completion-aggregator installed to collect profile data
```sh
python -m cProfile -o run_aggregator_service_master.profile ./manage.py lms run_aggregator_service
```

Install this branch to enable the caching.  Then browse another uncompleted section of the course.  Run profiling again with this branch installed.

```sh
python -m cProfile -o run_aggregator_service_branch.profile ./manage.py lms run_aggregator_service
```

**Reviewers:**
- [ ] @xitij2000 

FYI @bradenmacdonald 

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.

migrations, etc.


completion_aggregator/models.py
```diff
@@ -299,3 +299,15 @@ def __str__(self):
         if self.resolved:
             parts.append('*')
         return ''.join(parts)
+
+
+@python_2_unicode_compatible
+class CacheGroupInvalidation(models.Model):
```
xitij2000: Couldn't this be stored in the cache as well? Since the invalidation entry for a cache group will always happen after the entry itself, the original entry will also expire before the invalidation expires, so I don't think this will lead to data not being invalidated when it should. Or am I missing something about caching?

completion_aggregator/core.py
```diff
@@ -0,0 +1,322 @@
...
+        Also takes a modulestore instance.
+        """
+        self.user = user
+        self.course_key = course_key
+        self.raw_root_block = root_block
+        self.cache = UpdaterCache(self.user.id, self.course_key, self.raw_root_block)
+
+        cache_entry = self.cache.get()
+        if cache_entry:
+            self.course_blocks = cache_entry
```
xitij2000: ``self.root_block`` is no longer set if there is a cache entry, which causes an error to be thrown in ``calculate_updated_aggregators``:

```python
Traceback (most recent call last):
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/django/core/handlers/base.py", line 132, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/django/utils/decorators.py", line 145, in inner
    return func(*args, **kwargs)
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/django/views/decorators/csrf.py", line 58, in wrapped_view
    return view_func(*args, **kwargs)
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/django/views/generic/base.py", line 71, in view
    return self.dispatch(request, *args, **kwargs)
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/rest_framework/views.py", line 466, in dispatch
    response = self.handle_exception(exc)
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/rest_framework/views.py", line 463, in dispatch
    response = handler(request, *args, **kwargs)
  File "/edx/src/openedx-completion-aggregator/completion_aggregator/api/v1/views.py", line 408, in get
    ) for enrollment in paginated
  File "/edx/src/openedx-completion-aggregator/completion_aggregator/serializers.py", line 120, in __init__
    self.update_aggregators(aggregators or [], root_block=root_block, is_stale=is_stale)
  File "/edx/src/openedx-completion-aggregator/completion_aggregator/serializers.py", line 155, in update_aggregators
    force=True,
  File "/edx/src/openedx-completion-aggregator/completion_aggregator/core.py", line 293, in calculate_updated_aggregators
    return updater.calculate_updated_aggregators(changed_blocks, force)
  File "/edx/src/openedx-completion-aggregator/completion_aggregator/core.py", line 148, in calculate_updated_aggregators
    self.update_for_block(self.root_block, affected_aggregators, force)
AttributeError: 'AggregationUpdater' object has no attribute 'root_block'
```

completion_aggregator/core.py
```diff
@@ -0,0 +1,322 @@
+"""
```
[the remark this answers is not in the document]
xitij2000: No worries :-) wherever I saw some code moves I copy pasted the code in a diff program to see what had actually changed.

test_utils/test_app/migrations/0002_auto_20181217_1304.py
```diff
@@ -0,0 +1,68 @@
...
+        ),
+        migrations.AlterUniqueTogether(
+            name='courseaccessrole',
+            unique_together=set([('user', 'org', 'course_id', 'role')]),
+        ),
+        migrations.AlterUniqueTogether(
+            name='cohortmembership',
+            unique_together=set([('user', 'course_id')]),
+        ),
+    ]
```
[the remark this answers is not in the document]
jcdyer: I'll take a closer look.  Those are in the test app, (not in the aggregator proper).  They just got added when I did `makemigrations`, so I assumed someone had forgotten to add them when they updated the test app.

jcdyer:
I had some notes I'd written up that never got submitted as a review.  Just added them.

completion_aggregator/core.py
jcdyer: Oh yep.  I had some other information in the cache, and then I pulled it out.  Looks like I overdid it.  It's fixed now.  Please give it another look.

test_utils/test_app/migrations/0002_auto_20181217_1304.py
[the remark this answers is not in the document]
jcdyer: Yeah.  There's nothing particularly wrong with this, as it updates some stub models that are used in the test suite.  On the other hand, there's nothing particularly necessary about it, as the test db seems to get created just fine without it.  I lean towards leaving it in for "correctness," but I'm not strongly attached to it.

completion_aggregator/models.py
jcdyer: I'm not positive whether edx uses redis or memcache for caching, but if redis, there could be multiple cache nodes, and if one of them goes down, some data might get lost.  It might be okay, but I don't feel confident that the invalidation will always be present after the entry expires.  I'd rather have a durable database entry, just to be on the safe side.

xitij2000: Sure, that makes sense.

PR #50, opened 2018-12-20
MCKIN-9024 & BB-701: Make fetch course module completion list use POST instead of GET [WIP]

giovannicimolin:
**Description:** Describe in a couple of sentence what this PR adds

**JIRA:** Link to JIRA ticket

instructions.

**Testing instructions:**

1. Open page A
2. Do thing B
3. Expect C to happen
4. If D happend instead - check failed.

**Reviewers:**
- [ ] tag reviewer 

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.

migrations, etc.


completion_aggregator/api/v1/views.py
```diff
@@ -377,17 +387,17 @@ def get(self, request, course_key):
             # Use enrollments for the effective user
             enrollments = UserEnrollments(self.user).get_course_enrollments(course_key)
 
-        if 'user_ids' in request.query_params:
-            user_ids = (int(id) for id in request.query_params['user_ids'].split(','))
-            enrollments = enrollments.filter(user_id__in=user_ids)
+        user_ids = params.get('user_ids')
+        if user_ids:
+            enrollments = enrollments.filter(user_id__in=params['user_ids'])
```
Agrendalath: Nit: we could just use `user_ids` here.

completion_aggregator/api/v1/views.py
```diff
@@ -416,6 +426,28 @@ def get(self, request, course_key):
         )
         return paginator.get_paginated_response(serializer.data)
 
+    def get(self, request, course_key):
+        """
+        Handler for GET requests.
+        """
+        params = {}
+        if request.query_params.get('user_ids', None):
+            params['user_ids'] = (int(id) for id in re.split(r',|\.', request.query_params['user_ids']))
```
Agrendalath: We could simplify this regexp to `[,.]`. It will perform around 2x less steps this way.

completion_aggregator/api/v1/views.py
```diff
@@ -416,6 +426,28 @@ def get(self, request, course_key):
...
+
+        return self._parse_aggregator(course_key, params)
+
+    def post(self, request, course_key):
+        """
+        Handler for POST requests.
+        """
+        params = {}
+        params['user_ids'] = request.data.get('user_ids', None)
+        params['root_block'] = request.data.get('root_block', None)
```
Agrendalath: We could simplify it into:
```python
params = {
    'user_ids': request.data.get('user_ids'),
    'root_block': request.data.get('root_block'),
}
```

completion_aggregator/api/common.py
```diff
@@ -122,6 +122,9 @@ def user(self):
             return self._effective_user
 
         requested_username = self.request.GET.get('username')
+        if self.request.method == "POST":
+            requested_username = self.request.data.get('username', None)
```
Agrendalath: Isn't `None` default result for `get()` here?

completion_aggregator/api/v1/views.py
```diff
@@ -416,6 +426,28 @@ def get(self, request, course_key):
         )
         return paginator.get_paginated_response(serializer.data)
 
+    def get(self, request, course_key):
+        """
+        Handler for GET requests.
+        """
+        params = {}
+        if request.query_params.get('user_ids', None):
```
Agrendalath: Isn't `None` default result for `get()` here?

completion_aggregator/api/v1/views.py
```diff
@@ -416,6 +426,28 @@ def get(self, request, course_key):
         )
         return paginator.get_paginated_response(serializer.data)
 
+    def get(self, request, course_key):
+        """
+        Handler for GET requests.
+        """
+        params = {}
```
Agrendalath: Could we just initialize it like `params = {'root_block': request.query_params.get('root_block', None)}` here?

completion_aggregator/api/common.py
```diff
@@ -122,6 +122,9 @@ def user(self):
             return self._effective_user
 
         requested_username = self.request.GET.get('username')
```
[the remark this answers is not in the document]
giovannicimolin: Ok! I've made it like this to only get the username data when using `POST` but now I see that it doesn't make any sense to do that. I'll change it :+1:

completion_aggregator/api/v1/views.py
giovannicimolin: Woah! I've done some testing here and it's faster indeed. I'll make the changes. 
I'm not very experienced at regex so I went with what worked, but I'll read a little more about it.

Agrendalath: I personally liked [these](https://www.hackerrank.com/domains/regex) challenges for practicing `/regexp?/` :)

PR #51, opened 2019-01-16
MCKIN-9202: Exclude unregistered block types from completion.

jcdyer:
**Description:** Previously, the hourly aggregation task was failing whenever it encountered a course with an invalid block type.  This happened with historical courses that had run when different blocks were available.  Apparently at some time in the past, `image-explorer` was called `image_explorer`, and now any courses that contain an `image_explorer` block fail to aggregate completions, and the `StaleCompletion` records never get cleared out.  As a result, it tries to aggregate the same course every hour, and our client sees a few otherwise harmless errors in their NewRelic reporting.

This change should allow some of those records to be cleared out, and silence the errors.

**JIRA:** MCKIN-9209, BB-759

**Dependencies:** N/A

**Merge deadline:** None

On a solutions devstack, in the django admin, install this branch, then enable the waffle switch "completion.enable_completion_tracking".

**Testing instructions:**

1. Navigate through a few sections of a course, scrolling through each page, and reading the content.
2. Verify that completion_aggregator_stalecompletion records are created.
3. Rename one of the block types that is used in the course, by modifying `common/lib/xmodule/setup.py`: Under XMODULES, change (for example) `html = ...` to `htm = ...`.
4. Run `./manage.py lms --settings=devstack run_aggregator_service`.  Verify that values are added to the completion_aggregator_aggregator table, and that all entries in the completion_aggregator_stalecompletion table are marked "resolved."  The results will look off, because all of the "html" blocks are now invalid, but browsing the course on the site will likely be thoroughly broken anyway.  In the real use case, the now-invalid block type rarely shows up, and only in long-completed courses.

**Reviewers:**
- [ ] @giovannicimolin  

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.

migrations, etc.

### DIFF
--- a/completion_aggregator/management/commands/run_aggregator_service.py
+++ b/completion_aggregator/management/commands/run_aggregator_service.py
@@ -30,8 +30,8 @@ class Command(BaseCommand):
         """
         parser.add_argument(
             '--batch-size',
-            help='Maximum number of CourseModuleCompletions to migrate, per celery task. (default: 10000)',
-            default=10000,
+            help='Maximum number of StaleCompletions to process, per celery task. (default: 1000)',
+            default=1000,
             type=int,
         )
         parser.add_argument(
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -210,6 +210,9 @@ class CalculateUpdatedAggregatorsTestCase(TestCase):
             self.course_key.make_usage_key('html', 'course-chapter1-block2'),
             self.course_key.make_usage_key('html', 'course-chapter2-block1'),
             self.course_key.make_usage_key('html', 'course-chapter2-block2'),
+            # image_explorer is an unregistered block type, and should be
+            # treated as EXCLUDED from aggregation.
+            self.course_key.make_usage_key('image_explorer', 'course-chapter2-badblock'),
             self.course_key.make_usage_key('chapter', 'course-zeropossible'),
         ]
         patch = mock.patch('completion_aggregator.core.compat', StubCompat(self.blocks))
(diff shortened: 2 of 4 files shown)


tests/test_core.py
```diff
@@ -210,6 +210,9 @@ def setUp(self):
             self.course_key.make_usage_key('html', 'course-chapter1-block2'),
             self.course_key.make_usage_key('html', 'course-chapter2-block1'),
             self.course_key.make_usage_key('html', 'course-chapter2-block2'),
+            # image_explorer is an unregistered block type, and should be
+            # treated as EXCLUDED from aggregation.
+            self.course_key.make_usage_key('image_explorer', 'course-chapter2-badblock'),
```
jcdyer: This addition causes tests to fail when run against the current master branch.

completion_aggregator/management/commands/run_aggregator_service.py
```diff
@@ -30,8 +30,8 @@ def add_arguments(self, parser):
         """
         parser.add_argument(
             '--batch-size',
-            help='Maximum number of CourseModuleCompletions to migrate, per celery task. (default: 10000)',
-            default=10000,
+            help='Maximum number of StaleCompletions to process, per celery task. (default: 1000)',
+            default=1000,
```
jcdyer: Documentation cleanup, and modifying a default value that we had to override in practice, because the larger value caused slow queries and database lock contention.

jcdyer:
Thanks!  I've added a version bump for deployment purposes.  I will merge when tests pass again, and then ping you to approve a version bump PR on edx-solutions/edx-platform.

PR #52, opened 2019-01-17
YONK-1302: Reduce memory footprint of cache entries.

jcdyer:
**Description:** Reduce memory footprint of cache entries.

We do this by extracting only the information we need from the BlockStructure
object and storing it in simple data structures.

Also included:

* Removing redundant db queries.
* Only setting cache values when new values have been calculated (otherwise, just bump the timeout)

**JIRA:** YONK-1302, [BB-820](https://tasks.opencraft.com/browse/BB-820)

**Merge deadline:** ASAP

**Testing instructions:**

This is a performance refactor.  Verify that views that depend on completions still get accurate results, and that the size of cache entries is reduced by this operation.  

After visiting course pages in apros, check the size of the cached data.  To get the size of a cache entry, as stored in memcache, you need to first get the cache key, as follows:

```python
from util.memcache import safe_key

key = safe_key(
    'completion_aggregator.updater.{user_id}-{course_key}-{block_key}'.format(
        user_id=user_id, 
        course_key=course_key, 
        block_key=block_key or "COURSE", # Use "COURSE" to get the tree for the whole course.
    ),
    version=1,
    prefix='default',
)
```

Then use it to make a raw query to memcached, and check the size of the returned bytes.

```python
import socket
sock = socket.socket()
sock.connect(('localhost', 11211))
sock.send('get {}\n'.format(key))
response = sock.recv(2**30) # Get all the bytes.  If you think you didn't get all the bytes, try response += sock.recv(2**30) until the call blocks.
print(len(response))
```
The size of the response running this script against this branch should be multiple times smaller than running it against the master branch.

To clear the cache between runs do:

```python
from django.core.cache import cache
cache.clear()
```

**Reviewers:**
- [ ] @xitij2000 

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.

**Author notes/concerns:** 

* On the course I tested, the size of the memcache entry for a whole course was reduced from 500kb to 125kb, so 4x improvement.  Results over all courses in production may vary.
* The memory usage is better, I don't have a good sense of whether the improvement is sufficient, and no way to test it without seeing it in production.
* I'm also making an assumption that the cache entry size is the bottleneck.  Each call to the completion endpoint is only making a single call to set the cache value, so I think it's a reasonable assumption.
* I also spotted an excess database query which may be causing slowness.  See inline comment.
* The caching is here to deal with slowness in getting the course block structure in the first place.  If we can speed that up by not fetching student_view_data, caching may become unnecessary, but there is still an outstanding question of whether that will affect the correctness of results (if the structure of the course can ever depend on the content in student_view_data, perhaps for adaptive learning).

### DIFF
--- a/completion_aggregator/core.py
+++ b/completion_aggregator/core.py
@@ -5,6 +5,8 @@ This is the engine that takes BlockCompletion objects and
 converts them to Aggregators.
 """
 
+from __future__ import absolute_import, division, print_function, unicode_literals
+
 import logging
 from collections import namedtuple
 from datetime import datetime
@@ -64,6 +66,12 @@ class UpdaterCache(object):
         group = six.text_type(self.course_key)
         CacheGroup().set(group, self.cache_key, value, timeout=UPDATER_CACHE_TIMEOUT)
 
+    def touch(self):
+        """
+        Update the timeout for a given cache key.
+        """
+        CacheGroup().touch(self.cache_key, timeout=UPDATER_CACHE_TIMEOUT)
+
     @property
     def cache_key(self):
         """
@@ -74,6 +82,9 @@ class UpdaterCache(object):
             course_key=self.course_key,
             root_block=self.root_block or 'COURSE',
         )
+
+
+CourseBlocksEntry = namedtuple('CourseBlocksEntry', ['children', 'aggregators'])
 
 
 class AggregationUpdater(object):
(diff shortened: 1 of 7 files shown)


completion_aggregator/core.py
```diff
@@ -195,13 +217,10 @@ def update_for_aggregator(self, block, affected_aggregators, force):
         last_modified = OLD_DATETIME
 
         if block not in affected_aggregators:
-            try:
-                obj = Aggregator.objects.get(user=self.user, course_key=self.course_key, block_key=block)
-            except Aggregator.DoesNotExist:
-                pass
-            else:
+            obj = self.aggregators.get(block)
```
jcdyer: Here's where I save N database queries.  We already have all the aggregator blocks stored in AggregatorUpdater.__init__, so this is redundant.

tests/test_core.py
```diff
@@ -423,14 +423,13 @@ def test_multiple_partial_updates(self):
                 - auth_user (fetch user details)
                 - completion_aggregator_aggregator (user specific for specific course)
                 - completion_blockcompletion (user specific)
-                - completion_aggregator_aggregator (user specific for specific course and block)
```
jcdyer: This should have been a clue here that the query wasn't needed (Cf. L424)!

PR #56, opened 2019-02-24
[WIP] extend tests to python 3.6, 3.7 and django 2.1

dupainaulevain:
**Description:** extend tests to python 3.6, 3.7 and django 2.1


dupainaulevain:
It turns out the tests tests did not actually run on py35-django20: django [1.11 was installed instead](https://travis-ci.org/open-craft/openedx-completion-aggregator/jobs/495677967#L521). Maybe because setup.py had an **install_requires** of **Django>=1.8,<1.12** instead of **Django>=1.8,<2.2** ?

dupainaulevain:
This PR is a mess but has one merit: it showed that python3 support (hence django 2.0+ support) is blocked by https://github.com/edx/edx-celeryutils which does not currently support django 2.0 (but it is [tested with python 3.5](https://github.com/edx/edx-celeryutils/blob/master/.travis.yml#L8).

jcdyer:
@dupainaulevain Newer versions of pip let you specify required python versions for dependencies.  Something like:

    django-oauth-toolkit<x.y.z; python_version < 3
    django-oauth-toolkit; python_version >= 3

But don't quote me on the syntax.

I know pip-tools used to have problems with this syntax, but it probably has been fixed by now.  Then you wouldn't need entire separate sets of requirements.  You could also just separate out specific problematic requirements into another file, and pull them all together with pip-tools, depending on the python version.

jcdyer:
@dupainaulevain Tests should be running on python 3/django>=2.0 due to the last line of the `upgrade` target in Makefile.

dupainaulevain:
Ah right, sorry about that :-)

jcdyer:
No worries.  It's an obscure little bit of hackery that was in place to solve the old broken version of pip tools, but doesn't scale very well to handle multiple dependencies that are only compatible with a limited range of python releases.

PR #62, opened 2019-12-04
YONK-1535 - Updated code for edx-platform ironwood release compatibility

shafqatfarhan:
**Description:** Updated codebase to make it compatible with edx-platform ironwood release ensuring that the code can run across multiple releases

**JIRA:** https://openedx.atlassian.net/browse/YONK-1535

**Reviewers:**
- [ ] @xitij2000 

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.

### DIFF
--- a/completion_aggregator/api/common.py
+++ b/completion_aggregator/api/common.py
@@ -1,22 +1,14 @@
 """
 Common classes for api views
 """
-from django.contrib.auth import get_user_model
-from edx_rest_framework_extensions.auth.jwt.authentication import JwtAuthentication  # pylint: disable=import-error
-from edx_rest_framework_extensions.auth.session.authentication import SessionAuthenticationAllowInactiveUser  # pylint: disable=import-error
-from edx_rest_framework_extensions.paginators import NamespacedPageNumberPagination
 from rest_framework.exceptions import NotFound, ParseError, PermissionDenied
 from rest_framework.permissions import IsAuthenticated
+
+from django.contrib.auth import get_user_model
 
 from .. import compat
 from ..models import Aggregator
 from ..serializers import course_completion_serializer_factory, is_aggregation_name
-
-try:
-    from openedx.core.lib.api.authentication import OAuth2AuthenticationAllowInactiveUser
-except ImportError:
-    OAuth2AuthenticationAllowInactiveUser = object
-
 
 User = get_user_model()  # pylint: disable=invalid-name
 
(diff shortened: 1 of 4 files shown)


completion_aggregator/api/common.py
```diff
@@ -77,44 +85,20 @@ class CompletionViewMixin(object):
...
     permission_classes = (IsAuthenticated,)
+    _allowed_requested_fields = {'mean', 'username'}
     _effective_user = None
     _requested_user = None
 
     course_completion_serializer = None
     block_completion_serializer = None
 
-    @property
-    def authentication_classes(self):  # pragma: no cover
```
bradenmacdonald: I believe this can be kept as a method if renamed to [`get_authenticators()`](https://github.com/encode/django-rest-framework/blob/95d4843abeecea96754a147f4f2cca33e620ad09/rest_framework/views.py#L268-L272) (and not a `@property`)

bradenmacdonald:
Looks simple enough. @xitij2000 could you please coordinate a review?

xitij2000:
@bradenmacdonald  Sure will do. 
@shafqatfarhan I've scheduled this for review next sprint. However, I've had a quick look and noticed one major issue with the current changes. 

The code changes remove the compatibility code that ensures that the code can run across multiple releases. It does not seem necessary in this case. As commented above the authentication classes can be moved to method instead of the property. 

Here is an example from a different plugin that's comptaible from ginkgo to ironwood AFAIK:
```python
def get_authentication_classes():  # pragma: no cover
    """
    Allow users authenticated via OAuth2 or normal session authentication.
    """
    try:
        from openedx.core.lib.api.authentication import OAuth2AuthenticationAllowInactiveUser
        from openedx.core.lib.api.authentication import SessionAuthenticationAllowInactiveUser
    except ImportError:
        from edx_rest_framework_extensions.auth.session.authentication import SessionAuthenticationAllowInactiveUser

    try:
        from edx_rest_framework_extensions.auth.jwt.authentication import JwtAuthentication
    except ImportError:
        from edx_rest_framework_extensions.authentication import JwtAuthentication

    return [
        JwtAuthentication,
        OAuth2AuthenticationAllowInactiveUser,
        SessionAuthenticationAllowInactiveUser,
    ]
```
 
Also, please clean up the ticket description and remove the boilerplate.

shafqatfarhan:
@xitij2000 @bradenmacdonald Thank you for the review. I have updated the code according to the suggestions given and have used `get_authenticators()` method.

PR is ready for review.

xitij2000:
@shafqatfarhan It seems that this package is already broken on hawthorn since it upgrades edx-drf-extensions via completion. So I'm just testing on ironwood now. After that it's good to merge. 

@bradenmacdonald  Would it make sense bump to 2.0 and explicitly break compatibility with Hawthorn and below? 

Since you recently released completion 3.0 would you know if upgrading this to support completion 3.0 will require a similar break? If so the major version here can then be aligned with completion.

bradenmacdonald:
> @bradenmacdonald Would it make sense bump to 2.0 and explicitly break compatibility with Hawthorn and below?

Yes, that would be a good idea, and that the major version should be aligned with completion if possible, as you said.

> Since you recently released completion 3.0 would you know if upgrading this to support completion 3.0 will require a similar break?

It will, [3.0 introduces a breaking change to the python API](https://github.com/edx/completion/blob/master/CHANGELOG.rst#300---2019-10-22).

shafqatfarhan:
> 👍 This is good to merge.
> 
> * [x]  I tested this: tested on ironwood devstack
> * [x]  I read through the code
> 
> @shafqatfarhan I've just tested this on Ironwood and it's good to go.
> Could you please bump the version up to 2.0.0 to signify the break in compatibility? (Atlhough the actual break probably came with the previous PR)

@xitij2000 Done.

PR #64, opened 2020-03-02
Pinned edx-celeryutil function

nasirhjafri:
@xitij2000 Please review.


xitij2000:
@nasirhjafri  Could you include a version bump so that we can create a release?

PR #67, opened 2020-04-14
[BB-2296] Add locking mechanism to batch operations

Agrendalath:
**Description:** This adds cache-based locking to batch operations. This way we can safely schedule management command with cron without special restrictions for preventing race conditions in auto-scaling groups.
It also fixes `reaggregate_course` management command by using [course](https://github.com/edx/edx-platform/blob/open-release/ironwood.master/common/djangoapps/student/models.py#L1025) instead of `course_key`.

**JIRA:** [BB-2296](https://tasks.opencraft.com/browse/BB-2296)

**Testing instructions:**
1. Install aggregator from this branch.
1. Uncomment `batch.py:74` (deleting the lock).
1. Run aggregator `/edx/bin/manage.edxapp lms run_aggregator_service` (if you have `StaleCompletions` on your devstack, you should run it twice).
1. Run the command from the previous point again. You should get warning `Aggregation is already running. Exiting.`.
1. (Optionally) Remove the lock with `from django.conf import settings; cache.delete(settings.COMPLETION_AGGREGATOR_AGGREGATION_LOCK)` and run aggregation service again.

**Reviewers:**
- [x] @viadanna 

**Merge checklist:**
- [x] All reviewers approved
- [x] CI build is green
- [x] Version bumped
- [x] Changelog record added
- [x] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.


completion_aggregator/settings/common.py
```diff
@@ -16,3 +16,14 @@ def plugin_settings(settings):
...
     }
     settings.COMPLETION_AGGREGATOR_ASYNC_AGGREGATION = False
+
+    # Names of the batch operations locks
+    settings.COMPLETION_AGGREGATOR_AGGREGATION_LOCK = 'COMPLETION_AGGREGATOR_AGGREGATION_LOCK'
+    settings.COMPLETION_AGGREGATOR_CLEANUP_LOCK = 'COMPLETION_AGGREGATOR_CLEANUP_LOCK'
+
+    # Define how long should the locks be kept. They are released after completing the operation, so there are two
+    # possible scenarios for releasing the lock on timeout:
+    # 1. The management command takes more than 1800s - in this case you should set a higher limit for the lock.
```
viadanna, line 26: One alternative would be refreshing the timeout after aggregating each batch using `cache.touch`. This would in turn allow us to reduce the timeout. What do you think?

completion_aggregator/settings/common.py
```diff
@@ -16,3 +16,14 @@ def plugin_settings(settings):
...
+    # 2. The management command didn't exit successfully. You should check the logs to find out why.
+    settings.COMPLETION_AGGREGATOR_AGGREGATION_LOCK_TIMEOUT_SECONDS = 1800
+    settings.COMPLETION_AGGREGATOR_CLEANUP_LOCK_TIMEOUT_SECONDS = 900
```
viadanna, line 29: Since the initial run might take longer than subsequent ones, should we set this to a higher value? How often would this run, daily?

tests/test_batch.py
```diff
@@ -131,6 +139,21 @@ def test_with_no_blocks(mock_task, users):
     assert mock_task.call_count == 1
 
 
+@patch('completion_aggregator.tasks.aggregation_tasks.update_aggregators.apply_async')
+def test_lock(mock_task, users):
```
viadanna, line 143: Good call adding a test for this. Can you add one for the cleanup as well?

completion_aggregator/settings/common.py
Agrendalath: That would be possible, but it would require chaining batch of `update_aggregators` tasks and adding an extra step that releases the lock, because these are asynchronous celery tasks.

completion_aggregator/settings/common.py
Agrendalath: Currently once/twice per hour. The initial run will be manual (with `reaggregate_course` command).

tests/test_batch.py
Agrendalath: Sure, I'll add this :+1:

completion_aggregator/settings/common.py
viadanna: I meant refreshing the key inside [this loop](https://github.com/open-craft/openedx-completion-aggregator/blob/004337c442a49e44802fa3b48f04dfbe5749cdc4/completion_aggregator/batch.py#L82), but this can be done at a later time if this raises issues.

PR #77, opened 2020-07-13
MCKIN-21507 fixed an incorrect attribute

mudassir-hafeez:
**Description:** Fixed an incorrect attribute which was throwing an exception on New Relic. 

**JIRA:** https://edx-wiki.atlassian.net/browse/MCKIN-21507
https://openedx.atlassian.net/browse/YONK-1741

### DIFF
--- a/completion_aggregator/__init__.py
+++ b/completion_aggregator/__init__.py
@@ -4,6 +4,6 @@ an app that aggregates block level completion data for different block types for
 
 from __future__ import absolute_import, unicode_literals
 
-__version__ = '2.2.2'
+__version__ = '2.2.3'
 
 default_app_config = 'completion_aggregator.apps.CompletionAggregatorAppConfig'  # pylint: disable=invalid-name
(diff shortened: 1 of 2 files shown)


bradenmacdonald:
@mudassir-hafeez As I just mentioned on YONK-1741, I think that it may need to be `.update_or_create(group=group, defaults={"invalidated_at": timezone.now()})`

Otherwise, this will raise more errors - because the model defines `group` as unique.

mudassir-hafeez:
@bradenmacdonald thank you for reviewing it. Can you please review it once again?

bradenmacdonald:
@mudassir-hafeez I don't have enough context to sign off on this and I'm not able to test it now so @xitij2000 will review it for you soon.

xitij2000:
@mudassir-hafeez  Could you also add some tests? I will have a look at this soon.

xitij2000:
@mudassir-hafeez There is another bug here: https://github.com/open-craft/openedx-completion-aggregator/blob/master/completion_aggregator/cachegroup.py#L84
It should be `invalidate*d*_at__lt` instead of `invalidate_at__lt`

PR #80, opened 2020-11-11
[BB-2994] Improve this project's readme

bradenmacdonald:
This improves the completion aggregator readme to better explain what this project is, how it works, and why it's useful.

**Dependencies**: None

**Screenshots**: n/a

**Sandbox URL**: n/a

**Merge deadline**: None

**Testing instructions**:

1. Read the updated README [as rendered by GitHub](https://github.com/open-craft/openedx-completion-aggregator/blob/readme-update/README.rst).

**Author notes and concerns**:

None

**Reviewers**
- [x] @xitij2000 
- [x] @Agrendalath


README.rst
```diff
@@ -84,7 +99,7 @@ by each aggregator, and values are summed recursively from the course block on
 down.  Values for every node in the whole tree can be calculated in a single
 traversal.  These calculations can either be performed "read-only" (to get the
 latest data for each user), or "read-write" to store that data in the
-[`completion_aggregator.Aggregator` model](https://github.com/open-craft/openedx-completion-aggregator/blob/master/completion_aggregator/models.py#L199).
+`completion_aggregator.Aggregator model <https://github.com/open-craft/openedx-completion-aggregator/blob/a71ab4f077/completion_aggregator/models.py#L199>`_.
 
 During regular course interaction, a learner will calculate aggregations on the
 fly to get the latest information.  However, on-the-fly calculations are too
```
xitij2000, line 105: The documentation right after this mentions, `every hour or less` but this is not enforced by the code. Is this our recommendation? Should we have a recommendation?

README.rst
```diff
@@ -3,56 +3,71 @@ openedx-completion-aggregator
...
+
+The response will include a ``completion`` value between ``0`` and ``1``.
+
+However, what if you want to know the overall % completion of an entire course? ("Alex, you have completed 45% of Introduction to Statistics") Or what if you as an instructor want to get a report of how much of Section 1 every student in a course has completed? Those queries are either not possible or too slow using the APIs built in to the LMS and ``completion``.
+
+This Open edX plugin, ``openedx-completion-aggregator`` watches course activity and asynchronously updates database tables with "aggregate" completion data. "Aggregate" data means completion data summed up over all XBlocks into a course and aggregated at higher levels, like the subsection, section, and course level. The completion aggregator provides a REST API that can provide near-instant answers to queries such as:
+
+* What % complete are each of the courses that I'm enrolled in?
+* What % of each section in Course X have my students completed?
+* What is the average completion % among all enrolled students in a course?
```
xitij2000, line 27: Should we should clarify that aggregate queries like these will only be as accurate as the latest data based ont he last aggregation run?

README.rst
```diff
@@ -3,56 +3,71 @@ openedx-completion-aggregator
...
+
+    GET https://courses.opencraft.com/api/courses/v1/blocks/block-v1:OpenCraft+completion+demo+type@html+block@demo_block?username=MyUsername&requested_fields=completion
+
+The response will include a ``completion`` value between ``0`` and ``1``.
+
+However, what if you want to know the overall % completion of an entire course? ("Alex, you have completed 45% of Introduction to Statistics") Or what if you as an instructor want to get a report of how much of Section 1 every student in a course has completed? Those queries are either not possible or too slow using the APIs built in to the LMS and ``completion``.
+
+This Open edX plugin, ``openedx-completion-aggregator`` watches course activity and asynchronously updates database tables with "aggregate" completion data. "Aggregate" data means completion data summed up over all XBlocks into a course and aggregated at higher levels, like the subsection, section, and course level. The completion aggregator provides a REST API that can provide near-instant answers to queries such as:
```
xitij2000, line 23: While it does seem clear to me, should we mention that this comes with no UI?

README.rst
```diff
@@ -3,56 +3,71 @@ openedx-completion-aggregator
...
+            - subsection
+            - vertical
         ...
 
 
-3.  By default, completion is aggregated with each created or updated
-    BlockCompletion.  In most production instances, you will want to calculate
-    aggregations asynchronously.  To enable asynchronous calculation for your
-    installation, set the following in your lms.env.json file::
+3.  By default, completion is aggregated synchronously (with each created or updated BlockCompletion). While that is ideal for development, in most production instances, you will want to calculate aggregations asynchronously for better performance.  To enable asynchronous calculation for your installation, set the following in your ``lms.yml`` file::
```
xitij2000, line 54: I think it would be nice to clarify the difference between synchronous and asynchronous aggregation. I got a lot queries and there was a lot of confusion about how/when/what data is available and how accurate it is. 

i.e. With synchronous aggregation, each time a student completes a block, it will re-calculate the aggregates. You will always have the freshest results but at a huge performance cost. While with asynchronous aggregation, there will be a delay, since it will only aggregate results at periodic intervals. 
In both cases the single-user, single-course data endpoint will always return the freshest data.

README.rst
```diff
@@ -3,56 +3,71 @@ openedx-completion-aggregator
...
 
         ...
-        "COMPLETION_AGGREGATOR_ASYNC_AGGREGATION": true,
+        COMPLETION_AGGREGATOR_ASYNC_AGGREGATION: true
         ...
 
-    Then configure up a pair of cron jobs to run `./manage.py
-    run_aggregator_service` and `./manage.py run_aggregator_cleanup` as often
+    Then configure up a pair of cron jobs to run ``./manage.py
+    run_aggregator_service`` and ``./manage.py run_aggregator_cleanup`` as often
```
xitij2000, line 61: I think we should very briefly explain what these two do, and what running them more/less often means.

README.rst
```diff
@@ -3,56 +3,71 @@ openedx-completion-aggregator
...
+
+openedx-completion-aggregator uses the pluggable django app pattern to ease installation. To use in edx-platform, do the following:
+
+1.  Install the app into your virtualenv::
 
         $ pip install openedx-completion-aggregator
 
-2.  [Optional] You may override the set of registered aggregator block types in
-    your lms.env.json file::
+2.  [Optional] You may override the set of registered aggregator block types in your ``lms.yml`` file::
```
xitij2000, line 43: What will this do? Why would you use this?

xitij2000 (approved):
I think this greatly improves the non-technical parts of the documentation. I think there are still a few more questions we can answer here, but none of them very essential I feel. 

:+1:
- [na ] ~I tested this~
- [x] I read through the code
- [x] Includes documentation

README.rst
Agrendalath: Additionally to this it might be worth to make it **very** clear, that the synchronous aggregation can lead to a partial outage by causing deadlocks when users complete the blocks.
Unfortunately this is a real example…

README.rst
```diff
@@ -3,56 +3,71 @@ openedx-completion-aggregator
...
+3.  By default, completion is aggregated synchronously (with each created or updated BlockCompletion). While that is ideal for development, in most production instances, you will want to calculate aggregations asynchronously for better performance.  To enable asynchronous calculation for your installation, set the following in your ``lms.yml`` file::
 
         ...
-        "COMPLETION_AGGREGATOR_ASYNC_AGGREGATION": true,
+        COMPLETION_AGGREGATOR_ASYNC_AGGREGATION: true
         ...
 
-    Then configure up a pair of cron jobs to run `./manage.py
-    run_aggregator_service` and `./manage.py run_aggregator_cleanup` as often
+    Then configure up a pair of cron jobs to run ``./manage.py
```
Agrendalath, line 60: If the aggregator is installed on an existing instance, then it's sometimes desirable to fill "Aggregate" data for the existing courses. There is the [reaggregate_course](https://github.com/open-craft/openedx-completion-aggregator/blob/master/completion_aggregator/management/commands/reaggregate_course.py) management command, which prepares data that will be aggregated during the next `run_aggregator_service` run.
However that are a few points that might cause problems while doing this if you are running it for all courses on a big instance. People deploying this should be sure that they have:
1. Disabled cronjobs (if they have already enabled them). Otherwise these will clog Redis.
1. Enough computing power to run the aggregator for the first time. For our case we did a simulation by moving all data from prod to stage and measuring how long this will take. For one instance they were taking about 125h, so for the prod [we went with](https://tasks.opencraft.com/browse/BB-2296?focusedCommentId=148978&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-148978) disabling all existing Celery workers and spawning 3 new `m5.xlarge` instances running only workers with concurrency set to 4 (so 12 processes total).
1. Enough DB computing power. We have upgraded our RDS for this process, to be sure that the heavy parallelization won't make LMS unusable.
1. Enough DB storage (described [here](https://forum.opencraft.com/t/opencraft-ops-review/533/5)). "Aggregate" data is quite heavy - for us it added 20GB to the DB size during the initial run. We had RDS auto scaling enabled, but it throttled the instance significantly. Rescaling it manually fixed the issue, so it might be worth to scale DB upfront.

README.rst
Agrendalath: I just realized that we forgot to update docs in #67. This introduced locking mechanisms for batch aggregation and locking operations. Without this there was a possibility to creating an endless loop when aggregator ran too often - if it wouldn't be able to complete the whole operation in one cycle, it would start adding new task messages to Redis and they queue would be growing infinitely, as the tasks from the ongoing run would be duplicated.
So we added locks to prevent simultaneous runs, but if one run can take more than 30 minutes, the `COMPLETION_AGGREGATOR_AGGREGATION_LOCK_TIMEOUT_SECONDS` variable should be increased to prevent releasing the lock too early as a result of a timeout. Unfortunately the max duration of a single run will depend on the platform's usage, so it's not possible to provide sane values for everyone here.

README.rst
bradenmacdonald: Thanks, this is very valuable information and I've add this in to the README. Please take a look.

Is it `reaggregate_course` or `run_aggregator_service` that takes 125h? I assume the latter, but I thought you said you added a lock to prevent overlapping runs of it, so how did you use concurrent workers to speed it up?

README.rst
bradenmacdonald: I've removed that, I agree it's not useful for most people.

README.rst
bradenmacdonald: > I think we should very briefly explain what these two do, and what running them more/less often means.

OK, I took a stab at that.

> Then configure a pair of cron jobs to run ``./manage.py run_aggregator_service`` and ``./manage.py run_aggregator_cleanup`` as often as desired. (Start with hourly and daily, respectively, if you are unsure.) The ``run_aggregator_service`` task is what updates any aggregate completion data values that need to be updated since it was last run. The cleanup task deletes old database entries used to coordinate the aggregation updates, and which can build up over time but are no longer needed.

README.rst
bradenmacdonald: Excellent points. I've added a whole new section to the README about synchronous vs. asynchronous. Please check it out.

README.rst
bradenmacdonald: Sure, added a note.

> This service only provides data, via a REST API. There is no user interface.

README.rst
bradenmacdonald: Sure, added a sentence.

> (Note that answers to these "aggregate" questions may be slightly out of date, because they are computed asynchronously. How often they are updated is configurable.)

README.rst
bradenmacdonald: I guess it would ideally just be a process that constantly runs again as soon as it's finished, unless there are no stale completions. But yeah, I think it really depends on the usage of the instance. How about if I just say "e.g. hourly"

For now I've just said

> (e.g. every hour, but this is configurable)

@Agrendalath Thanks for that information - I've added this:

> 3. A lock is used to prevent overlapping runs of ``run_aggregator_service``. If you have a lot of data on your instance such that a single run of ``run_aggregator_service`` takes longer than 30 minutes, you must configure ``COMPLETION_AGGREGATOR_AGGREGATION_LOCK_TIMEOUT_SECONDS`` to a larger value to prevent releasing the lock too early as a result of a timeout.

bradenmacdonald:
@xitij2000 @Agrendalath  Thanks for the very helpful review! I've made updates based on your advice.

README.rst
Agrendalath: @bradenmacdonald, it's `run_aggregator_service`. `reaggregate_course` didn't need more than a few minutes.

> I thought you said you added a lock to prevent overlapping runs of it, so how did you use concurrent workers to speed it up?

Oh, I now remember the specific use case we had in mind here.
The reason we implemented this is because we have this specified in Ansible as:
```yaml
EDXAPP_ADDITIONAL_CRON_JOBS:
  - name: "run aggregator service"
    user: "{{ edxapp_user }}"
    job: ". {{ edxapp_app_dir }}/edxapp_env && {{ edxapp_venv_bin }}/python {{ edxapp_code_dir }}/manage.py lms run_aggregator_service --settings={{ edxapp_settings }} >/dev/null 2>&1"
    minute: "*/30"
    hour: "*"
    day: "*"
  - name: "run aggregator cleanup"
    user: "{{ edxapp_user }}"
    job: ". {{ edxapp_app_dir }}/edxapp_env && {{ edxapp_venv_bin }}/python {{ edxapp_code_dir }}/manage.py lms run_aggregator_cleanup --settings={{ edxapp_settings }} >/dev/null 2>&1"
    minute: "15"
    hour: "*"
    day: "*"
```
Therefore this job is present on each instance in the auto scaling group. This would fire all of them simultaneously, so we are using Redis cache locks here (which use atomic operations) to be sure that only one instance is able to run this.

The lock is only for scheduling Celery tasks, not for running them. We didn't create the locks to protect ourselves from the infinite loops, but this can be an issue if you run the `run_aggregator_service` too often while having tons of new block completions. This is not a very likely scenario, but it's technically possible. This could potentially be mitigated by marking `StaleCompletion`s as `resolved` once they are scheduled for the Celery task (or by adding a `scheduled` field to this model), but I'm not fully aware whether this would not cause any side effects.
Sorry for the confusion on this!

README.rst
```diff
@@ -26,6 +26,22 @@ This Open edX plugin, ``openedx-completion-aggregator`` watches course activity
...
+
+* This service only provides data, via a REST API. There is no user interface.
+* On production instances, the answers to these "aggregate" questions may be slightly out of date, because they are computed asynchronously (see below). How often they are updated is configurable.
+
+Synchronous vs. Asynchronous calculations
+-----------------------------------------
+
+openedx-completion-aggregator operates in one of two modes: synchronous or asynchronous.
+
+With synchronous aggregation, each time a student completes a block, the aggregator code will re-calculate the aggregate completion values immediately. You will always have the freshest results from this API, but at a huge performance cost. Synchronous aggregation is only for development purposes and is not suitable for production. Synchronous aggregation can cause deadlocks when users complete XBlocks, leading to a partial outage of the LMS.
```
Agrendalath, line 39: @bradenmacdonald, please make the last sentence **bolded**.

README.rst
```diff
@@ -26,6 +26,22 @@ This Open edX plugin, ``openedx-completion-aggregator`` watches course activity
...
+
+With asynchronous aggregation, the aggregator code will re-calculate the aggregate completion values asynchronously, at periodic intervals (e.g. every hour). How often the update can and should be run depends on many factors - you will have to experiment and find what works best and what is possible for your specific Open edX installation.
```
Agrendalath, line 41: @bradenmacdonald, we could also add a warning that running this too often can clog the tasks queue, which might require manual intervention.

README.rst
```diff
@@ -40,26 +56,17 @@ openedx-completion-aggregator uses the pluggable django app pattern to ease inst
...
 
-    Then configure up a pair of cron jobs to run ``./manage.py
-    run_aggregator_service`` and ``./manage.py run_aggregator_cleanup`` as often
-    as desired.
+4. If the aggregator is installed on an existing instance, then it's sometimes desirable to fill "Aggregate" data for the existing courses. There is the ``reaggregate_course`` management command, which prepares data that will be aggregated during the next ``run_aggregator_service`` run. However, the process of aggregating data for existing courses can place extremely high loads on both your celery workers and your MySQL database, so on large instances this process must be planned with great care. For starters, we recommend you disable any associated cronjobs, scale up your celery worker pool significantly, and scale up your database cluster and storage. To give you some idea, on one Open edX instance this process took 125 hours.
```
Agrendalath, line 69: Side note: it would have taken 125h if we had used 1 worker with its default configuration (1 process). When we ran 12 processes, it took around 11h.

Agrendalath (approved):
@bradenmacdonald, I left some smaller comments, but LGTM and it probably won't require another round of reviews.

:+1:
* [x]  I tested this: read the docs
* [x]  I read through the code
* [x]  I checked for accessibility issues: n/a
* [x]  Includes documentation

README.rst
```diff
@@ -3,56 +3,78 @@ openedx-completion-aggregator
...
-completion data for different block types for Open edX.
+openedx-completion-aggregator is a Django app that aggregates block level completion data for different block types for Open edX.
 
-Overview
---------
+What does that mean?
 
-openedx-completion-aggregator uses the pluggable django app pattern to
-ease installation.  To use in edx-platform, do the following:
+A standard Open edX installation can track the completion of individual XBlocks in a course, which is done using the `completion library <https://github.com/edx/completion#completion>`_. This completion tracking is what powers the green checkmarks shown in the course outline and course navigation as the learner completes each unit in the course:
```
Agrendalath, line 10: We might want to change [this line](https://github.com/open-craft/openedx-completion-aggregator/blob/master/tox.ini#L5) to something like `ignore=D001` to make `doc8` happy in the CI.

README.rst
bradenmacdonald: @Agrendalath 
I see, thanks.

Ok, so does it happen sometimes that `run_aggregator_service` (which just enqueues the tasks, not actually processes them) can take longer than 30 minutes? I just want to ensure this wording is still correct:

> A lock is used to prevent overlapping runs of run_aggregator_service. If you have a lot of data on your instance such that a single run of run_aggregator_service takes longer than 30 minutes, you must configure COMPLETION_AGGREGATOR_AGGREGATION_LOCK_TIMEOUT_SECONDS to a larger value to prevent releasing the lock too early as a result of a timeout.

README.rst
bradenmacdonald: OK. I'll just remove that sentence anyways.

README.rst
Agrendalath: @bradenmacdonald, good point, `run_aggregator_service` shouldn't take that much time in the regular usage (not counting its manual run after `reaggregate_course`). We can remove this part and add it in case we need some locks shared between celery tasks in the future.

xitij2000 (approved):

PR #89, opened 2021-10-18
[SE-4473] Add progress bar views

pkulkark:
**Description:** 
This PR adds new views that display progress bars for course-level and chapter-level progress.
This requires the changes from https://github.com/open-craft/edx-platform/pull/431 to be present.

**JIRA:** [SE-4473](https://tasks.opencraft.com/browse/SE-4473)

**Dependencies:** https://github.com/open-craft/edx-platform/pull/431 

**Testing instructions:**

1. Once you have a edx devstack setup with changes from https://github.com/open-craft/edx-platform/pull/431, install and enable completion tracking.
2. Pip install this django app and checkout the PR branch.
3. Make sure the feature flag "SHOW_PROGRESS_BAR" is set to True on your devstack.
4. Start a new course in LMS and check that the two progress bars are displayed and updated correctly.

**Reviewers:**
- [ ]

### DIFF
--- a/completion_aggregator/views.py
+++ b/completion_aggregator/views.py
@@ -25,14 +25,15 @@ class CompletionProgressBarView(LoginRequiredMixin, TemplateView):
         Fetch progress and render the template.
         """
         completion_percentage = 0
+        new_req = request.GET.copy()
+        new_req['username'] = request.user.username
         if chapter_id is not None:
-            new_req = request.GET.copy()
             new_req['requested_fields'] = "chapter"
-            request.GET = new_req
+        request.GET = new_req
         completion_resp = CompletionDetailView.as_view()(request, course_key).data
         if completion_resp:
             results = completion_resp.get('results')
-            user_completion_percentage = self._get_user_completion(request.user.username, chapter_id, results)
+            user_completion_percentage = self._get_user_completion(chapter_id, results)
 
             if user_completion_percentage:
                 completion_percentage = user_completion_percentage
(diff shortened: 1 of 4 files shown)


completion_aggregator/views.py
```diff
@@ -0,0 +1,54 @@
...
+from django.views.decorators.clickjacking import xframe_options_exempt
+from django.views.generic import TemplateView
+from xblockutils.resources import ResourceLoader
+
+from .api.v1.views import CompletionDetailView
+
+loader = ResourceLoader(__name__)
+
+
+class CompletionProgressBarView(LoginRequiredMixin, TemplateView):
```
gabor-boros, line 18: How about the following refactor of this logic?

```python
class CompletionProgressBarView(LoginRequiredMixin, TemplateView):
    """
    View to display the progress bar of a student in a course
    """
    def __get_user_completion(username, chapter_id, results):
        """
        Return the user completion percentage, using the completion response.

        In case the user completion cannot be returned as a result of missing user completion, we return None,
        indicating its absence.
        """
        user_completion = next(filter(results, lambda r: r.get('username') == username), None)

        # No completion returned, hence we cannot get the percentage either
        # Indicate no user completion by returning None
        if not user_completion:
            return None

        chapters = user_completion['chapter']
        chapter = next(filter(chapters, lambda c: c['block_key'].split('@')[-1] == chapter_id), None)

        completion_kind = chapter if chapter else user_completion
        return float(completion_kind['completion']['percent']) * 100

    @xframe_options_exempt
    def get(self, request, course_key, chapter_id=None):
        """
        Fetch progress and render the template.
        """
        completion_percentage = 0
        completion_resp = CompletionDetailView.as_view()(request, course_key).data

        if chapter_id is not None:
            new_req = request.GET.copy()
            new_req['requested_fields'] = "chapter"
            request.GET = new_req

        if completion_resp:
            results = completion_resp.get('results')
            user_completion_percentage = self.__get_user_completion(request.user.username, chapter_id, results)

            if user_completion_percentage:
                completion_percentage = user_completion_percentage

        template = 'chapter_completion_progress_bar.html' if chapter_id is not None else 'completion_progress_bar.html'
        return render(request, template, {
            'completion_percentage': completion_percentage,
        })
```

gabor-boros: Also, could you please add tests for the view?

gabor-boros (changes requested):

PR #278, opened 2026-01-11
chore: add support for Python 3.11 + 3.12 and Django 4.2 + 5.2

Agrendalath:
**Description:**
**Note: please review this commit-by-commit, as it gives the context of the changes.**

This:
* Drops support for Python 2, Python 3.8, and Django 3.2.
* Adds support for Python 3.11 + 3.12 and Django 4.2 + 5.2.
* Fixes the ``stats`` API endpoint.
* Fixes the `ENROLLMENT_TRACK_UPDATED` signal.
* Removes unused compat imports.

**Merge deadline:** None

**Testing instructions:**

1. Check that the CI is passing (ignore Codecov, as it's irrelevant in this case).

*Private-ref: [BB-10290](https://tasks.opencraft.com/browse/BB-10290)*


requirements/constraints.txt
```diff
@@ -10,9 +10,3 @@
 
 # Common constraints for edx repos
 -c https://raw.githubusercontent.com/edx/edx-lint/master/edx_lint/files/common_constraints.txt
```
BryanttV:
```suggestion
-c https://raw.githubusercontent.com/openedx/edx-lint/master/edx_lint/files/common_constraints.txt
```

BryanttV (commented):
Hi @Agrendalath! Thanks for this. Over the last month, we have been using this package with this [draft PR](https://github.com/eduNEXT/openedx-completion-aggregator/pull/1), which we planned to publish upstream once the test error was fixed. I see that you did manage to fix it!

On my side, I’ve been testing this branch, and everything works correctly, thank you very much. If you think it’s appropriate, I would recommend using `django-upgrade` to update the code to Django 4.2. This is the command I used:

```sh
git ls-files -z -- '*.py' | xargs -0r django-upgrade --target-version 4.2
```

tecoholic (approved):
@Agrendalath :+1:

- [x] I tested this: Verified that the CI is passing as expected.
- [x] I read through the code
- [ ] I checked for accessibility issues - N/A
- [x] Includes documentation

CHANGELOG.rst
```diff
@@ -14,6 +14,13 @@ Change Log
 Unreleased
 ~~~~~~~~~~
 
+[4.2.1] - 2026-01-27
```
tecoholic: @Agrendalath `4.3.0` would be better right? Moving support from Django 3.2 to 4.2 feel like a bigger change than a patch fix.

